PR #30, opened 2018-01-25
Feature: managed aliases

Davidmattei:
Environment local and external should only use **1** index, but we want to manage aliases with multiple indexes (indices).

On the environment page we can now view/create/edit and delete managed aliases. Managed aliases are stored in 'managed_alias' table. It's also possible to align managed aliases on their edit page.

Added indexes counter to the unreferenced aliases overview. Disabled the attach function for unreferenced aliases with multiple indexes. And also added the delete function for unreferenced aliases.

### DIFF
--- a/Entity/ManagedAlias.php
+++ b/Entity/ManagedAlias.php
@@ -1,0 +1,246 @@
+<?php
+
+namespace EMS\CoreBundle\Entity;
+
+use EMS\CoreBundle\Validator\Constraints as EMSAssert;
+use Doctrine\ORM\Mapping as ORM;
+use Doctrine\ORM\Mapping\PrePersist;
+use Symfony\Bridge\Doctrine\Validator\Constraints\UniqueEntity;
+
+/**
+ * Managed Alias
+ *
+ * @ORM\Table(name="managed_alias")
+ * @ORM\Entity(repositoryClass="EMS\CoreBundle\Repository\ManagedAliasRepository")
+ * @ORM\HasLifecycleCallbacks()
+ * 
+ * @UniqueEntity(fields={"name"}, message="Name already exists!")
+ */
+class ManagedAlias
+{
+    /**
+     * @var int
+     *
+     * @ORM\Column(name="id", type="integer")
+     * @ORM\Id
+     * @ORM\GeneratedValue(strategy="AUTO")
+     */
+    private $id;
+    
+    /**
+     * @var string
+     * 
+     * @EMSAssert\AliasName()
+     * 
+     * @ORM\Column(name="name", type="string", length=255, unique=true)
+     */
+    private $name;
+    
+    /**
+     * @var string
+     * @ORM\Column(name="alias", type="string", length=255, unique=true)
+     */
+    private $alias;
+
+    /**
+     * @var \DateTime
+     *
+     * @ORM\Column(name="created", type="datetime")
+     */
+    private $created;
+
+    /**
+     * @var \DateTime
+     *
+     * @ORM\Column(name="modified", type="datetime")
+     */
+    private $modified;
+    
+    /**
+     * @var array
+     */
+    private $indexes = [];
+    
+    /**
+     * @var int
+     */
+    private $total;
+  
+    /**
+     * @var string
+     *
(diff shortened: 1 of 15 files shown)


Resources/views/elements/post-button.html.twig
```diff
@@ -1,6 +1,6 @@
 {% trans_default_domain 'EMSCoreBundle' %}{% if message is defined %}
 <div class="btn-group">
-  <button type="button" class="btn {% if class is defined %}{% else %}btn-sm{% endif %} btn-danger dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+  <button type="button" class="btn {% if class is defined %}{{ class }}{% else %}btn-sm btn-danger{% endif %} dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
```
dameert: To be verified with MDK, I think POST actions should always be btn-danger

ems-project: Agree, but it fix a bug

Repository/EnvironmentRepository.php
```diff
@@ -22,6 +22,17 @@ public function findBy(array $criteria, array $orderBy = null, $limit = null, $o
...
     }
+    
+    /**
+     * @return array
+     */
+    public function findAllAliases()
+    {
+        $qb = $this->createQueryBuilder('e', 'e.alias');
+        $qb->select('e.alias, e.name, e.managed');
+
```
ems-project: I'm confuse, does the environment entity as an alias field or an aliases field

Resources/DoctrineMigrations/Version20180129102508.php
```diff
@@ -0,0 +1,34 @@
+<?php
```
ems-project: The migration script should be generated in the local-user-bundle (for mysql, pgsql and sqlite)

Resources/views/environment/index.html.twig
```diff
@@ -75,7 +75,7 @@
 							<th>Action</th>
 						</tr>
 						{% for environment in environments %}
-							{{ 'Environmen\'s info'|debug({name: environment.name}) }}
+							{{ "Environmen's info"|debug({name: environment.name}) }}
```
ems-project: This is an interesting update. I must be from the past: https://stackoverflow.com/questions/3446216/what-is-the-difference-between-single-quoted-and-double-quoted-strings-in-php

Resources/views/environment/index.html.twig
```diff
@@ -93,7 +93,7 @@
 						                </span>		
 									</td>
 									<td>{{ environment.alias }}</td>
-									<td>{{ environment.index }}</td>
```
ems-project: see other comments

Repository/EnvironmentRepository.php
Davidmattei: Yes the environment entity has the following columns: id, name, alias, color, managed, ... The alias is the environment name prefixed with the instance_id parameter.

Entity/ManagedAlias.php
```diff
@@ -0,0 +1,246 @@
...
+    
+    /**
+     * @var array
+     */
+    private $indexes = [];
+    
+    /**
+     * @var int
+     */
+    private $total;
```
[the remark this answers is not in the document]
Davidmattei: It's the total documents for the managed alias.
$this->client->count(['index' => 'aliasName'])

Resources/views/environment/index.html.twig
Davidmattei: Fixed

Resources/views/elements/post-button.html.twig
dameert: Ok, so no longer the red look obligation for Post buttons?

Resources/DoctrineMigrations/Version20180129102508.php
Davidmattei: https://github.com/ems-project/EMSLocalUserBundle/pull/2/files

PR #41, opened 2018-03-30
Fix title attribute for object pickers

Davidmattei:
Set the correct title attribute on the select options.
Now select2 will use the html value as title on load.

### DIFF
--- a/Form/Field/ObjectChoiceListItem.php
+++ b/Form/Field/ObjectChoiceListItem.php
@@ -6,6 +6,7 @@ use EMS\CoreBundle\Entity\ContentType;
 class ObjectChoiceListItem {
 
 	private $label;
+	private $title;
 	private $value;
 	private $group;
 	private $color;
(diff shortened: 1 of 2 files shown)


Form/Field/ObjectChoiceListItem.php
```diff
@@ -26,9 +27,11 @@ public function __construct(array &$object, ContentType $contentType){
 			$this->label = '<i class="'.(null !== $contentType->getIcon()?$contentType->getIcon():'fa fa-question').'" data-ouuid="'.$this->value.'"></i>&nbsp;&nbsp;';
 			if(null !== $contentType->getLabelField() && isset($object['_source'][$contentType->getLabelField()])){
 				$this->label .= $object['_source'][$contentType->getLabelField()];
+				$this->title = $object['_source'][$contentType->getLabelField()];
```
dameert: I think this title attribute still has all the html markup (e.g <i class=".." oouid=".."></i>Realtitle)

PR #47, opened 2018-04-06
debug migrate for collection of file

IsaMic:
Adapt migration to find all field (no matter how depth) of given name field in raw of a revision


theus77:
Ok to me. just that the scope of the command has been extended. And the command should be renamed to 'ems:revisions:index-file-field'

And we also need to add an option to index only if the fields has not been indexed yet.

PR #58, opened 2018-04-24
improvement: migration command

Davidmattei:
- Don't flush inside loop document, flush before continue scrolling
- Progress should advance by scrollsize

This improve the migration time by +/- 30%


Davidmattei:
Not good!

PR #61, opened 2018-04-27
cron option for recompute

dameert:
I think we should require at least the option "--keep-align" or "--no-align". As in, you have to specify one of those when you run.

Because, when you run without --keep-align by accident, you will have all last revisions in your default environment. So a rerun with --keep-align will not repair the alignment.


dameert:
image
There seems to be an issue with the date field using the recompute. The last date is used for all revisions in the revisions tab..

Davidmattei:
Very good pull request! 

Greetings

David

PR #107, opened 2019-02-11
fix: env extra field available in CT template

Davidmattei:
example env extra info:
{"nl": "https://example-nl.test","fr": "https://example-fr.test"}

example template:
{{ environment.extra|json_decode.nl }}


Davidmattei:
Work around use get_environment filter: 
{{ environment.name|**get_environment**.extra|json_decode.nl }}

PR #117, opened 2019-03-17
Stan lvl 0 config and 84 error fixes

dameert:
Too many errors, first effort:
From 446 to 381 errors @ level 0.


dameert:
Notable changes that really need a review are commited with the prefix "review:"

Controller/ContentManagement/EnvironmentController.php
```diff
@@ -502,12 +502,12 @@ public function addAction(Request $request)
 
     /**
      * Edit environement (name and color). It's not allowed to update the elasticsearch alias.
-     * @param unknown $id
+     * @param integer $id
```
Davidmattei: int

dameert: I agree, but the whole project contains "integer" issues. This will be taken care off when passing to "php 7.2 strict typing" (I have 360 other issues for now on level 0)

PR #123, opened 2019-03-19
phpstan swap service container for specific data service

dameert:


theus77:
Have you tried this refactoring? If I added this it was because I faced some circle references. I'll try on my own tonight

dameert:
I did not try it. But circle references seem strange in this case. If you have an error, post it here and I will look into it.

dameert:
I tried and php bin/console c:cl passes without errors.

theus77:
I tried, it works now. Good news.

PR #125, opened 2019-03-19
phpstan fix notification class issues

dameert:


Service/NotificationService.php
```diff
@@ -154,15 +157,9 @@ public function setStatus(Notification $notification, $status, $level = 'notice'
     
     /**
      * Call addNotification when click on a request
-     *
-     * @param unknown $template
-     * @param unknown $revision
      */
-    public function addNotification($templateId, $revision, $environment)
+    public function addNotification(string $templateId, Revision $revision, Environment $environment)
```
dameert: code calling this function is compatible with strong typing (string, Revision & Environment)

theus77: I think that a template id id an integer, not a string. Or it's a string but converted to a integer somewhere

Twig/AppExtension.php
```diff
@@ -1060,6 +1060,11 @@ public function searchesList($username)
         return $searches;
     }
 
+    /**
+     * PHPStan complains about this, we should really consider using the {% dump %} functionality and Symfony StackTrace.
+     * And dump got added in twig 1.5:
+     * https://twig.symfony.com/doc/1.x/functions/dump.html
+     */
     public function dump($object)
```
theus77: Agree, it can be flagged as deprecated

Service/NotificationService.php
dameert: @theus77 
Looks like there is a problem on line 170:
$template = $repository->findOneById($templateId);

The function findOneById does not exist in the repositories.

dameert: Nvm i found the magic in __call of EntityRepository (doctrine)

Twig/AppExtension.php
dameert: I will do this in another pull request

PR #128, opened 2019-03-20
phpstan fix reverseViewTransform method

dameert:


Form/DataField/DataFieldType.php
```diff
@@ -74,7 +74,7 @@ public function postFinalizeTreatment($type, $id, DataField $dataField, $previou
      *
      * http://symfony.com/doc/current/form/data_transformers.html#about-model-and-view-transformers
      *
-     * @param unknown $data
+     * @param array|string $data
      * @param FieldType $fieldType
```
theus77: It can be what ever we want: int, double, string, .... a json object?

dameert: Ok I see it now, based on the viewTranform method:
array|null|string|integer|float

PR #138, opened 2019-03-25
Crypto js

theus77:


Resources/views/base.html.twig
```diff
@@ -13,7 +13,9 @@ scratch. This page gets rid of all links and provides the needed markup only.
     <meta content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no" name="viewport">
 
     <!-- stylesheets -->
-    {%- if theme_color %}
+    <link rel="stylesheet" href="{{ asset('bundles/emscore/css/app.bundle.css') }}">
+    {%- if theme_color in  ['blue2'] %}
+        {% dump('bundles/emscore/css/'~theme_color~'.bundle.css') %}
```
dameert: Reviewing per commit, you should rebase and force push on your local branch if newer commit is solving this trailing dump. That way the error is not visible during review.

dameert:
Which are the priorities to migrate from bower to webpack?
Ckeditor would be nice to further enhance the style definitions

PR #140, opened 2019-03-29
Bye bower

theus77:


Resources/assets/js/core.js
```diff
@@ -1,14 +1,17 @@
 import FileUploader from "../FileUploader/FileUploader";
 
 const $ = require('jquery');
-const jQuery = require('jquery');
+
+
+// const jQuery = require('jquery');
```
dameert: remove it if we don't need it. Git history always has the trace

PR #173, opened 2019-04-15
Copy files command

Davidmattei:
Requires: https://github.com/ems-project/EMSCommonBundle/pull/26

Example: 

php bin/console ems:copy:files c:\external-files\pdf --symlink

### DIFF
--- a/Command/CopyFilesCommand.php
+++ b/Command/CopyFilesCommand.php
@@ -1,0 +1,92 @@
+<?php
+
+namespace EMS\CoreBundle\Command;
+
+use EMS\CommonBundle\Command\CommandInterface;
+use EMS\CommonBundle\Storage\Service\FileSystemStorage;
+use EMS\CoreBundle\Service\FileService;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+use Symfony\Component\Finder\Finder;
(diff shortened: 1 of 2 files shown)


Command/CopyFilesCommand.php
```diff
@@ -0,0 +1,92 @@
...
+        parent::__construct();
+        $this->fileService = $fileService;
+        $this->storageFolder = $storageFoler;
+    }
+
+    protected function configure()
+    {
+        parent::configure();
+        $this
+            ->setDescription('Copy or symlink files too ems storage folder')
```
theus77: Typo tooo

Command/CopyFilesCommand.php
```diff
@@ -0,0 +1,92 @@
...
+        $style->title(sprintf('%s files from %s\%s to %s', ($symlink ? 'Symlink' : 'Copy'), $source, $pattern, $this->storageFolder));
+
+        $files = Finder::create()->in($source)->files()->name($pattern);
+
+        $progress = $style->createProgressBar($files->count());
+        $progress->start();
+
+        foreach ($files as $file) {
+            /** @var $file SplFileInfo */
+            $hash = \sha1_file($file->getPathname());
```
theus77: Please use the hash_file of the storage service (the hash algorithm might not be sha1)

Command/CopyFilesCommand.php
```diff
@@ -0,0 +1,92 @@
...
+            if ($fileSystem->head($hash)) {
+                $progress->advance();
+                continue;
+            }
+
+            if ($symlink) {
+                $fileSystem->symlink($hash, $file->getPathname());
+            } else {
+                $fileSystem->create($hash, $file->getPathname());
+            }
```
theus77: Thé core storage service is bypassed, so the file aren’t registered in the asset table. Is it a good idea? Not sure.

PR #176, opened 2019-04-29
Phpstan

dameert:
PHPStan fixes (level 2)

Review commit per commit, I tried to work per file so I can easily cherry pick if certain changes are not approved. Thx


Service/DataService.php
```diff
@@ -237,16 +238,17 @@ public function propagateDataToComputedField(FormInterface $form, array& $object
         return $this->propagateDataToComputedFieldRecursive($form, $objectArray, $contentType, $type, $ouuid, $migration, $objectArray, '');
     }
 
-    private function propagateDataToComputedFieldRecursive(FormInterface $form, array& $objectArray, ContentType $contentType, $type, $ouuid, $migration, &$parent, $path)
+    private function propagateDataToComputedFieldRecursive(Form $form, array& $objectArray, ContentType $contentType, $type, $ouuid, $migration, &$parent, $path)
```
theus77: Why does the Form class is better than the Interface? does the function use non-interface function?

Form/Field/EnvironmentPickerType.php
```diff
@@ -77,7 +68,7 @@ public function configureOptions(OptionsResolver $resolver)
                     'data-live-search' => false
             ],
             'choice_attr' => function ($category, $key, $index) {
-                /** @var \EMS\CoreBundle\Form\DataField\DataFieldType $dataFieldType */
+                /** @var Environment $dataFieldType */
                 $dataFieldType = $this->choices[$index];
```
theus77: We should also rename the var

Service/DataService.php
dameert: Yes ->getIterator on line 334
(bullet 3 in my commit message https://github.com/ems-project/EMSCoreBundle/pull/176/commits/22071a276125ec3b8e6fcdb94c78e3fc60cc20bb )

Form/Field/EnvironmentPickerType.php
dameert: I agree with renaming, and we should add an "instanceOf" check to prevent errors.

Service/DataService.php
dameert: I found a better solution ! see last commit

Controller/WysiwygController.php
```diff
@@ -152,7 +153,7 @@ public function editStylesSetAction(Request $request, WysiwygStylesSet $stylesSe
         $form->handleRequest($request);
         
         if ($form->isSubmitted()) {
-            if ($form->get('remove') && $form->get('remove')->isClicked()) {
+            if ($form->get('remove') instanceof ClickableInterface  && $form->get('remove')->isClicked()) {
```
theus77: $removeForm = $form->get('remove');
if ( $removeForm instanceof ....

Controller/WysiwygController.php
```diff
@@ -181,11 +182,11 @@ public function editStylesSetAction(Request $request, WysiwygStylesSet $stylesSe
      */
     public function editProfileAction(Request $request, WysiwygProfile $profile)
     {
-        $form= $this->createForm(WysiwygProfileType::class, $profile);
+        $form = $this->createForm(WysiwygProfileType::class, $profile);
         $form->handleRequest($request);
 
         if ($form->isSubmitted()) {
-            if ($form->get('remove') && $form->get('remove')->isClicked()) {
+            if ($form->get('remove') instanceof ClickableInterface && $form->get('remove')->isClicked()) {
```
theus77: $removeForm = $form->get('remove');
if ( $removeForm instanceof ....

Service/DataService.php
theus77: Ok to me

PR #186, opened 2019-05-10
Do not detach the user object when verifying circles state

dameert:


dameert:
The bug is caused by the "lockRevision" function in the DataService Class.
Line 153:             if (!$this->appTwig->inMyCircles($revision->getRawData()[$revision->getContentType()->getCirclesField()])) {

$this->appTwig->inMyCircles detatches the user, and so ORM is trying to re-create it.

dameert:
To be verified, does this break calls to "in_my_circles" from twig?

dameert:
> To be verified, does this break calls to "in_my_circles" from twig?

Tested it on the "draft in progress" view, seems to be working ok !

PR #188, opened 2019-05-14
Stan lvl three

dameert:


Entity/UploadedAsset.php
```diff
@@ -261,7 +261,7 @@ public function getType()
     /**
      * Set user
      *
-     * @param string $user
+     * @param User $user
```
theus77: Nope it's a string, there is non hard link between the ems core data model and a user entity. Always a soft username string.

Entity/View.php
```diff
@@ -58,7 +58,7 @@ class View
     private $icon;
     
     /**
-     * @var string
```
theus77: Nope it's a string, there is non hard link between the ems core data model and a user entity. Always a soft username string.

Entity/UploadedAsset.php
theus77: we should probably rename this variable to username

dameert: Ok I Will take care of this. I see Indeed that doctrine mapping is string. I Will change the @var of the property

Entity/View.php
dameert: This code is about the $options array. That is a json_array in doctrine. In php this is an array.
This is proven by it's usage on many places: 
$view->getOptions()['targetField']

PR #197, opened 2019-05-21
add createFromJson and exportToJson function

IsaMic:


Service/ContentTypeService.php
```diff
@@ -409,4 +416,60 @@ public function getAllTypes()
...
+            'json'
+        );
+        $contentType->setEnvironment($environment);
+        $contentType->setActive(false);
+        $contentType->setDirty(true);
+        $contentType->getFieldType()->updateAncestorReferences($contentType, null);
+        $contentType->setOrderKey($contentTypeRepository->maxOrderKey() + 1);
+        
+        $this->persist($contentType);
+        return $contentType->getId();
```
dameert: I think that your should return the $contentType object, if another function only needs the id, its that other function that should call the getId() function.

Service/ContentTypeService.php
```diff
@@ -409,4 +416,60 @@ public function getAllTypes()
...
+        $contentType->setModified(null);
+        $contentType->getFieldType()->removeCircularReference();
+        $contentType->setEnvironment(null);
+        
+        //Serialize the CT
+        $encoders = array(new JsonEncoder());
+        $normalizers = array(new JsonNormalizer());
+        $serializer = new Serializer($normalizers, $encoders);
+        $jsonContent = $serializer->serialize($contentType, 'json');
+        $response = new Response($jsonContent);
```
dameert: Respones are only usefull for Controllers that render a web page (Response).
This function should return the $jsonContent, and the controller using this function can then generate the response.

theus77:
I'm confused, those functionalities already exist, no?

Capture d’écran 2019-05-21 à 20 19 36

dameert:
Yes inside a controller so not useable for our command. That is why we isolate them in the service. We decided to not refactor the controller in this first step, and only do What we need to get the command to work. (That is why I mention the deprecated statement that Should be added in the controller so we do not forget that this isolation is alleady done).

Controller/ContentManagement/ContentTypeController.php
```diff
@@ -228,27 +228,15 @@ public function addAction(Request $request)
...
-                        "EMS\CoreBundle\Entity\ContentType",
-                        'json'
-                    );
+                    
+                    $contentType = $this->getContentTypeService()->readJson($file, $environment);
+                    
                     $contentType->setName($name);
                     $contentType->setSingularName($singularName);
                     $contentType->setPluralName($pluralName);
                     $contentType->setEnvironment($environment);
```
dameert: Environment is allready passed to the readJson function, I think this call to setEnvironment can be removed!

Service/ContentTypeService.php
```diff
@@ -15,6 +15,14 @@
 use EMS\CoreBundle\Entity\Environment;
 use EMS\CoreBundle\Entity\FieldType;
 use Symfony\Component\Translation\TranslatorInterface;
+use Symfony\Component\Serializer\Encoder\JsonEncoder;
+use Symfony\Component\Serializer\Serializer;
+use EMS\CoreBundle\Entity\Helper\JsonNormalizer;
+use Symfony\Component\HttpFoundation\Response;
```
dameert: clean up unused dependencies

Service/ContentTypeService.php
```diff
@@ -409,4 +417,57 @@ public function getAllTypes()
         $this->loadEnvironment();
         return implode(',', array_keys($this->contentTypeArrayByName));
     }
+    
+    public function readJson(File $jsonFile, Environment $environment): ContentType
```
dameert: I think a good name could be "initFromJson", as we are not only reading the json, but also setting the environment.
We are initializing a contentType based on a json.

Service/ContentTypeService.php
```diff
@@ -409,4 +417,57 @@ public function getAllTypes()
...
+        $contentType = $serializer->deserialize(
+            $fileContent,
+            "EMS\CoreBundle\Entity\ContentType",
+            'json'
+        );
+        $contentType->setEnvironment($environment);
+        return $contentType;
+    }
+    
+    public function createContentType(ContentType $contentType): ContentType
```
dameert: shorter named methods are better (if they are still clear on their purpose). We are in the ContentTypeService, so "create" is a good function name

dameert: After reading the code, I see we change the content type that is provided, and we persist it.
To reflect the business of this function I would vote for "persistAsNew"

dameert:
The controller actions are simplified in the "serializer" branch. This should be reflected here before merging to master. (I'll take care of it)

dameert:
https://github.com/ems-project/EMSCoreBundle/commit/f7da03c262914e6eb9f8ec5ce6b7f9f451fa638d allready includes these changes

PR #198, opened 2019-05-25
Export view pdf support

theus77:


Form/View/ExportViewType.php
```diff
@@ -117,9 +138,32 @@ public function getBlockPrefix()
      */
     public function generateResponse(View $view, Request $request)
     {
-        $response = new Response();
         $parameters = $this->getParameters($view, $this->formFactory, $request);
-        
+
+        if (isset($view->getOptions()['export_type']) or $view->getOptions()['export_type'] === 'dompdf') {
```
dameert: Preference for || instead of 'or' (convention?), as there is no precedence issue here

Form/View/ExportViewType.php
```diff
@@ -117,9 +138,32 @@ public function getBlockPrefix()
...
+            // instantiate and use the dompdf class
+            $dompdf = new Dompdf();
+            $dompdf->loadHtml($parameters['render']);
+
+            // (Optional) Setup the paper size and orientation
+            $dompdf->setPaper(
+                (isset($view->getOptions()['pdf_size']) and $view->getOptions()['pdf_size']) ? $view->getOptions()['pdf_size'] : 'A4',
```
dameert: preference for && instead of 'and' (convention?), as there is no precendence issue here

Form/View/ExportViewType.php
```diff
@@ -117,9 +138,32 @@ public function getBlockPrefix()
...
+                (isset($view->getOptions()['pdf_orientation']) and $view->getOptions()['pdf_orientation']) ? $view->getOptions()['pdf_orientation'] : 'portrait'
+            );
+
+            // Render the HTML as PDF
+            $dompdf->render();
+
+            // Output the generated PDF to Browser
+            $dompdf->stream($parameters['filename'] ?? "document.pdf", [
+                'compress' => 1,
+                'Attachment' => ( isset($view->getOptions()['disposition'])  && $view->getOptions()['disposition']  === 'attachment')?1:0,
```
dameert: there is a "space" too much before the && charachtesr, and before the === charachters. And seeing the && here gives more reasen to follow this convention in the other if checks.

Form/View/ExportViewType.php
```diff
@@ -117,9 +138,32 @@ public function getBlockPrefix()
...
+            ]);
+            exit;
```
dameert: This function is doing multiple things, and to terminate in case of PDF generation, this exit is present?
I think the function should be split. Creation of a PDF should happen in it's own function, an exit statement should not be needed.

Form/View/ExportViewType.php
theus77: tha'is an expensive space ;-)

Form/View/ExportViewType.php
theus77: Agree, the pdf generation should be relocated in a service

PR #200, opened 2019-05-27
Serializer

dameert:
Serialization as a class responsability, removing the normalizer


Entity/SingleTypeIndex.php
```diff
@@ -205,4 +207,20 @@ public function getContentType()
...
+     * @return mixed data which can be serialized by <b>json_encode</b>,
+     * which is a value of any type other than a resource.
+     * @since 5.4.0
+     */
+    public function jsonSerialize()
+    {
+        $json = new JsonClass(get_object_vars($this), __CLASS__);
+        $json->removeProperty('id');
+        $json->removeProperty('environment');
+
```
theus77: So, what the meaning of breaking this link? Is it not better to call the reconnect command?

dameert: When exporting and importing json files between projects, we cannot guarantee that the same environment names are used in both projects. This is also done for the contenttype json.

When importing, the environment to use should be specified (as we allready require for content types)

dameert: We cannot

> I'm not comfortable with the links between entities

Do you mean the "extends JsonDeserializable" ? Why are you not comfortable with that link?

dameert:
> I'm not comfortable with the links between entities

Do you mean the "extends JsonDeserializable" ? Why are you not comfortable with that link?

PR #206, opened 2019-06-04
Better reindex command

theus77:


dameert:
Make sure that phpstan passes at level 3 (see pull request #207 )

PR #210, opened 2019-06-09
Tika debugs

theus77:


Service/AssetExtractorService.php
```diff
@@ -62,6 +63,10 @@ public function __construct(RestClientService $rest, LoggerInterface $logger, Re
      */
     private function getTikaWrapper() : ?TikaWrapper
     {
+        if (! empty($this->tikaServer)) {
```
dameert: Not the responsability of this function. The warmup should handle this test.

PR #236, opened 2019-08-22
comply to psr12

dameert:


theus77:
One of the most borriest merge request I ever seen... but thanks a lot for this.

PR #238, opened 2019-08-28
refactor: import _self as macro for twig 2.11

dameert:
"import _self as" no longer works from inside another macro.
Neither does calling macro's dynamically.

However we use import _self as at the base level in EMSCommonBundle, which is still functioning. So CommonBundle is backward compatible with twig 2.10! (important for our skeleton projects in production). But also operates fine on twig 2.11 (required for EMSCoreBundle form this commit on)

Requires:
https://github.com/ems-project/EMSCommonBundle/pull/47

### DIFF
--- a/Controller/ElasticsearchController.php
+++ b/Controller/ElasticsearchController.php
@@ -720,6 +720,7 @@ class ElasticsearchController extends AppController
 
                 $searchArray['filters'] = null;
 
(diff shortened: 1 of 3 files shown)


dameert:
I think this change should be released as a new minor vers: 1.11.x

PR #241, opened 2019-09-11
Debug postpocess sen email

theus77:


Service/EnvironmentService.php
```diff
@@ -172,7 +172,7 @@ public function getById($id)
     public function getManagedEnvironement()
     {
         /** @var Environment $environment*/
```
dameert: Think this comment is not necessary if you type hint in the function

PR #245, opened 2019-11-10
Min should and match phrase support

theus77:


Entity/Form/Search.php
```diff
@@ -353,4 +360,22 @@ public function setContentType($contentType)
         $this->contentType = $contentType;
         return $this;
     }
+
+    /**
+     * @return int
```
dameert: Redundant comment as return type is defined allready

Entity/Form/Search.php
```diff
@@ -353,4 +360,22 @@ public function setContentType($contentType)
...
+     */
+    public function getMinimumShouldMatch(): int
+    {
+        return $this->minimumShouldMatch;
+    }
+
+    /**
+     * @param int $minimumShouldMatch
+     * @return Search
```
dameert: Redundant comment block

Entity/Form/Search.php
theus77: Solved, but I would like to find a way to disable itnfrom php storm? https://www.jetbrains.com/help/phpstorm/settings-file-and-code-templates.html?

dameert: I think the very last point in this article might help https://www.jetbrains.com/help/phpstorm/phpdoc-comments.html

PR #249, opened 2019-11-20
 Bugfix: removed condition to fix query filter

laurensmertens:
Needed for https://jira.smals.be/browse/WEBCHAMBER-525


theus77:
You can define a default search per content type by saving a search form than define it as default for the content type. This query will be used by CK editor.

PR #250, opened 2019-11-26
Export command (in order to replace the actual export in another PR)

theus77:


dameert:
Import/export commands should live in the makerbundle. The architecture is allready set up and contenttype is implemented

Command/ExportCommand.php
```diff
@@ -0,0 +1,145 @@
...
+use EMS\CoreBundle\Service\Mapping;
+use Monolog\Logger;
+use Symfony\Component\Console\Helper\ProgressBar;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Form\FormFactoryInterface;
+
+class ExportCommand extends EmsCommand
```
dameert: Export what? Name should be specific.

Command/ExportCommand.php
```diff
@@ -0,0 +1,145 @@
...
+        $progress->start();
+
+        while (isset($arrayElasticsearchIndex['hits']['hits']) && count($arrayElasticsearchIndex['hits']['hits']) > 0) {
+            foreach ($arrayElasticsearchIndex["hits"]["hits"] as $index => $value) {
+                $progress->advance();
+            }
+
+            $scroll_id = $arrayElasticsearchIndex['_scroll_id'];
+            $arrayElasticsearchIndex = $this->client->scroll([
+                "scroll_id" => $scroll_id,  //...using our previously obtained _scroll_id
```
dameert: Copy pasted comments?

Command/ExportCommand.php
```diff
@@ -0,0 +1,145 @@
...
+        if (! $contentType instanceof ContentType) {
+            $output->writeln(sprintf("WARNING: Content type named %s not found", $contentType));
+            return null;
+        }
+        $index = $input->getArgument('index');
+        if ($index === null) {
+            $index = $contentType->getEnvironment()->getAlias();
+        }
+
+        $arrayElasticsearchIndex = $this->client->search([
```
dameert: The business logic of this command should be abstracted away. Use yield in your abstraction if you want to handle the progress bar in the command.

Entity/ContentType.php
```diff
@@ -1768,4 +1775,15 @@ protected function deserializeProperty(string $name, $value)
...
+
+    public function getBusinessIdField(): ?string
+    {
+        return $this->businessIdField;
+    }
+
+    public function setBusinessIdField(string $businessIdField): ContentType
+    {
+        $this->businessIdField = $businessIdField;
+        return $this;
```
dameert: This should be discussed but in general a setter should return void.

theus77:
The point of this WIP merge request is to migrate the actual export into a command as now it can handle large export.

dameert:
But no code is removed? (So no migration of code?) if it is a command to export data/config I still think all my comments apply, the name of the command should be specific and be part of the existing maker bundle)

Command/ExportCommand.php
theus77: Renamed to ExportDocumentsCommand

Entity/ContentType.php
theus77: Until now all setter in the core are fluent

Command/ExportCommand.php
theus77: cleaned

Command/ExportCommand.php
theus77: Fair point (I'll do it like this in the future)

PR #251, opened 2019-11-27
Ready: Update mapping indexer

theus77:


laurensmertens:
All good

dameert:
General advice (to be defined together): Independent commits should be in a separate pr. The smaller a pr the better.

PR #252, opened 2019-11-28
Allow copy_to mapping option for ChoiceFieldType

theus77:


theus77:
The travis fails will be resolved after the merge (fixed in master)

PR #255, opened 2019-12-16
Better job experience

theus77:


Controller/TwigElementsController.php
```diff
@@ -43,6 +43,16 @@ public function sideMenuAction(AssetExtractorService $assetExtractorService, Cli
         );
     }
 
+    public function jobsAction(string $username, JobService $jobService)
```
dameert: The use of Action in the function name is deprecated since Symfony 4. Might be better to comply for new functions? And even transform in a separate pull request the other functions of this file?

Service/JobService.php
```diff
@@ -67,6 +67,16 @@ public function clean()
         $this->em->flush();
     }
 
+    public function findByUser($user)
```
dameert: Type hinting?

theus77: Agree fixed

Controller/TwigElementsController.php
theus77: Agree but the twig controller function append an action (see the base.html.twig)

PR #256, opened 2019-12-16
The paging size is hard coded to 30 everywere in the JS app

theus77:


dameert:
Would it not be better to resolve the hard coding in JS? And default the parameter to 30 instead of removing it (which I think you forgot in the config file?)

theus77:
@dameert in fact the paging_size parameter is still used to define the page size of the search result pages. As I do not see the point to synchronize the paging size in the select2 drop down menu with that one, I have considered to hard code it like that

PR #257, opened 2019-12-18
Remove orphaned indexes command

laurensmertens:
Added the delete orphaned indexes command and fixed a typo in ContentManagement/IndexController.php


Command/DeleteOrphanIndexesCommand.php
```diff
@@ -0,0 +1,46 @@
...
+    {
+        $this->setName('ems:delete:orphans')
+            ->setDescription('Removes all orphan indexes');
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $this->aliasService->build();
+        foreach ($this->aliasService->getOrphanIndexes() as $index) {
+            $this->client->indices()->delete([
```
theus77: It might be useful to surround this by a try catch if it’s not possible to delete an index. With a log error in the catch. So even if it’s not possible to delete one orphan index the command can continue its cleaning

Command/DeleteOrphanIndexesCommand.php
```diff
@@ -0,0 +1,46 @@
...
+                'index' => $index['name'],
+            ]);
+            $this->logger->notice('log.index.delete_orphan_index', [
```
theus77: Please add a translation in the English translation file

Command/DeleteOrphanIndexesCommand.php
theus77: And move the try catch and the delete into a private function deleteOrpheanIndex? To avoid nested code structure

laurensmertens: OK fixed it like this

Command/DeleteOrphanIndexesCommand.php
laurensmertens: It was already translated but doesn't appear in the console, I used $output->writeln like in the other commands now which is more clear.

laurensmertens: Fixed it back to the translatable log

Command/DeleteOrphanIndexesCommand.php
```diff
@@ -0,0 +1,58 @@
...
+    {
+        try {
+            $this->client->indices()->delete([
+                'index' => $index['name'],
+            ]);
+            $this->logger->notice('log.index.delete_orphan_index', [
+                'index_name' => $index['name'],
+            ]);
+        } catch (Missing404Exception $e) {
+            $this->logger->notice('log.index.index_not_found', [
```
theus77: This is at least a  warning or an error but not "just" a notice

Command/DeleteOrphanIndexesCommand.php
```diff
@@ -36,19 +36,23 @@ protected function execute(InputInterface $input, OutputInterface $output)
     {
         $this->aliasService->build();
         foreach ($this->aliasService->getOrphanIndexes() as $index) {
-            $this->deleteOrphanIndex($index, $output);
+            $this->deleteOrphanIndex($index);
         }
     }
 
-    private function deleteOrphanIndex($index, OutputInterface $output)
```
theus77: Type hint for the funciton parameters is missing

Command/DeleteOrphanIndexesCommand.php
laurensmertens: Good point, fixed

Command/DeleteOrphanIndexesCommand.php
laurensmertens: Also fixed

laurensmertens:
@theus77 Can we merge this?

Command/DeleteOrphanIndexesCommand.php
```diff
@@ -0,0 +1,58 @@
...
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $this->aliasService->build();
+        foreach ($this->aliasService->getOrphanIndexes() as $index) {
+            $this->deleteOrphanIndex($index);
+        }
+    }
+
+    private function deleteOrphanIndex(array $index)
```
dameert: This function does exactly the same as deleteOrphansIndexesAction in the Controller. Maybe it's time to refactor this in a IndexService ? 

Also this would improve DI : $client = $this->getElasticsearch(); is deprecated (use DI instead)

Command/DeleteOrphanIndexesCommand.php
```diff
@@ -0,0 +1,58 @@
...
+    public function __construct(LoggerInterface $logger, Client $client, AliasService $aliasService)
+    {
+        $this->logger = $logger;
+        $this->client = $client;
+        $this->aliasService = $aliasService;
+        parent::__construct($logger, $client);
+    }
+
+    protected function configure()
+    {
```
dameert: assigning command name is now done using a variable:

https://symfony.com/doc/current/console.html
protected static $defaultName = 'app:create-user';

dameert: But do not do this when extending from EMSCommand ! because it will get overriden (to be reviewed in a later stade)

laurensmertens: How do you mean? DeleteOrphanIndexesCommand will overwrite EMSCommand anyway won't it?

dameert: protected static $defaultName = 'app:create-user'; Would not work because setName is called in the parent. So setting a defaultName in this extended class would not work I think (to be tested?)

laurensmertens: This works, instances of classes always use their own properties instead of parent properties, even when using parent methods. I also tested this to make sure.

dameert:
Hey Laurens, you forgot to check phpcs, the failing issues can be fixed running phpcbf !

Command/DeleteOrphanIndexesCommand.php
```diff
@@ -0,0 +1,40 @@
...
+
+class DeleteOrphanIndexesCommand extends EmsCommand
+{
+    /** @var LoggerInterface */
+    protected $logger;
+    /** @var Client */
+    protected $client;
+
+    protected static $defaultName = 'ems:delete:orphans';
+    /**
```
dameert: use one line @var notation

Elasticsearch/Indexer.php
```diff
@@ -91,7 +91,7 @@ public function atomicSwitch(string $alias, string $newIndex, string $removeRege
         array_map([$this, 'delete'], $delete);
     }
 
-    private function getClient(): Client
+    public function getClient(): Client
```
dameert: I am not sure why you make this functino public? This is not called in your other changes? Revert this back to private.

Service/IndexService.php
```diff
@@ -0,0 +1,42 @@
...
+
+use Elasticsearch\Client;
+use Psr\Log\LoggerInterface;
+
+final class IndexService
+{
+    /** @var AliasService */
+    private $aliasService;
+    /** @var Client */
+    private $elasticSearchClient;
```
dameert: I think we always use 'client' as variable name for this one.

Service/IndexService.php
```diff
@@ -0,0 +1,42 @@
...
+    public function __construct(AliasService $aliasService, Client $elasticSearchClient, LoggerInterface $logger)
+    {
+        $this->aliasService = $aliasService;
+        $this->elasticSearchClient = $elasticSearchClient;
+        $this->logger = $logger;
+    }
+
+    public function deleteOrphanIndexes(): void
+    {
+        $this->aliasService->build();
```
dameert: Why is this build() call needed? In the old code it was never called in the indexController function deleteOrphansIndexesAction()?

Command/DeleteOrphanIndexesCommand.php
laurensmertens: This was fixed last night normally

Elasticsearch/Indexer.php
laurensmertens: Indeed, I used it before creating a dedicated service. Fixed now.

Service/IndexService.php
laurensmertens: OK

Service/IndexService.php
laurensmertens: If I don't build it up front, getOrphanIndexes() returns an empty array.

dameert: So this orphaned index action was never working before ?

dameert: Nvm, it was "getElasticSearch()" that invoked the build.

I don't realy like this external call to ->build(), but that is out of scope for this change.

PR #259, opened 2020-01-13
Public access for get indexer client

laurensmertens:


dameert:
The indexer should not expose getClient. It is not a factory for the client! It’s a helper class for Index functionality!

dameert:
+ why repeat all commits of your previous PR? Did you create a branch from a non master branch? Rebase with master pls!

laurensmertens:
The append script needs to this client to get the correct index.

PR #260, opened 2020-01-14
Added getAliases to indexer

laurensmertens:
Used in the append option for import

### DIFF
--- a/Elasticsearch/Indexer.php
+++ b/Elasticsearch/Indexer.php
@@ -91,7 +91,7 @@ class Indexer
         array_map([$this, 'delete'], $delete);
     }
 
-    public function getAliases(string $indexName): array
+    public function getAliasesByIndex(string $indexName): array
     {
         return $this->getClient()->indices()->getAliases(['index' => $indexName]);
     }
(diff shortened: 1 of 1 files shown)


Elasticsearch/Indexer.php
```diff
@@ -95,4 +95,9 @@ private function getClient(): Client
     {
         return $this->factory->fromConfig($this->options);
     }
+
```
dameert: Put public functions before private functions

Elasticsearch/Indexer.php
```diff
@@ -95,4 +95,9 @@ private function getClient(): Client
...
+    public function getAliases($params = array()): array
```
dameert: Do we want to pass an array with Params or only the index name as a string? Maybe we should convert this to “getAliasByIndexName” to make the call easier?

laurensmertens: OK good idea

Elasticsearch/Indexer.php
laurensmertens: Done

PR #265, opened 2020-02-06
Export with command

theus77:


Controller/ContentManagement/JobController.php
```diff
@@ -56,8 +56,17 @@ public function jobStatusAction(Job $job)
         $theme = new Theme();
         $converter = new AnsiToHtmlConverter($theme);
 
+        $status = preg_replace_callback(
```
dameert: Logic should not be part of the controller. Can this be done in a data object?

Controller/ElasticsearchController.php
```diff
@@ -612,6 +615,37 @@ public function searchApiAction(Request $request)
         ]);
     }
 
+    /**
+     * @param Request $request
```
dameert: Remove docblock info that is already type hinted in the function definition

Controller/ElasticsearchController.php
```diff
@@ -612,6 +615,37 @@ public function searchApiAction(Request $request)
...
+     */
+    public function exportAction(Request $request, JobService $jobService, ContentType $contentType)
+    {
+        $exportDocuments = new ExportDocuments($contentType, $this->generateUrl('emsco_search_export', ['contentType' => $contentType]), '{}');
+        $form = $this->createForm(ExportDocumentsType::class, $exportDocuments);
+        $form->handleRequest($request);
+
+        /** @var ExportDocuments */
+        $exportDocuments = $form->getData();
+        $command = sprintf(
```
dameert: Why run a command from the controller?? This is a strange behaviour. The controller and commands are interfaces to underlying services. Abstract the logic from the command to reuse it here. I think this apply to all jobs allready defined?

Controller/ContentManagement/JobController.php
theus77: This is a hack which one detects url within the last line of the job ouput. This is a view stuff but I wont take care of it in the twig and I don' t need it anywhere else (nor create a twig fucntion). But ideally the AnsiToHmltConverter should support the [Console Hyperlink](https://symfony.com/blog/new-in-symfony-4-3-console-hyperlinks)

Controller/ElasticsearchController.php
theus77: That's how works ems asynchronous calls Todqy. I agree, it's an hack. I'm looking for a better solution. Nevertheless there is commit in this pull request which ones improve this functionality (flush output in DB on every writeln and infinite time limit).

Controller/ElasticsearchController.php
theus77: Removed as recommended

Controller/ContentManagement/JobController.php
dameert: Okay but this is technical debt. I would prefer a correct implementation because it’s not clear what is happening (=will waste dev time if someone passes this function)

Controller/ElasticsearchController.php
```diff
@@ -647,14 +643,6 @@ public function exportAction(Request $request, JobService $jobService, ContentTy
     }
 
     /**
-     * @param Request $request
-     * @return RedirectResponse|Response
-     * @throws Throwable
```
dameert: Keep throws. And return type as it is not defined in the function call

Controller/ElasticsearchController.php
```diff
@@ -616,10 +616,6 @@ public function searchApiAction(Request $request)
     }
 
     /**
-     * @param Request $request
-     *
-     * @param JobService $jobService
-     * @return RedirectResponse
      * @Route("/search/export/{contentType}", name="emsco_search_export", methods={"POST"})
```
dameert: Keep return as it is not defined in the function definition

Controller/ElasticsearchController.php
dameert: A message broker like redis should be considered

Controller/ElasticsearchController.php
theus77: Rollbacked

Controller/ElasticsearchController.php
theus77: Agree

Controller/ElasticsearchController.php
theus77: Rollbacked

Controller/ElasticsearchController.php
theus77: Rollbacked

Controller/ContentManagement/JobController.php
theus77: refactored in the common

Controller/ContentManagement/JobController.php
```diff
@@ -130,15 +131,16 @@ public function startJobAction(Job $job, Request $request, JobService $jobServic
             return new JsonResponse('job already done');
         }
 
+        set_time_limit(0);
```
dameert: Set time limit 0 Will not override php-fpm time limitations! And platform specific limitations imposed by the hosting environment.

PR #273, opened 2020-02-13
Feature / HTML styles remover

coppee:


tests/ContentTransformer/HtmlStylesRemoverTest.php
```diff
@@ -0,0 +1,276 @@
...
+    public function testRemoveSimpleInline()
+    {
+        $input = <<<HTML
+<p>Lorem ipsum dolor <span class="removable-style-newWord">sit amet</span>, consectetur adipiscing elit.</p>
+HTML;
+        $output = <<<HTML
+<p>Lorem ipsum dolor sit amet, consectetur adipiscing elit.</p>
+HTML;
+
+        $htmlStylesRemover = new HtmlStylesRemover();
```
dameert: Isn't there a "setup" function that can be used for this? (because every test case is using this line of code).

tests/ContentTransformer/HtmlStylesRemoverTest.php
```diff
@@ -0,0 +1,276 @@
...
+namespace EMS\CoreBundle\Tests\ContentTransformer;
+
+use Symfony\Bundle\FrameworkBundle\Test\WebTestCase;
+use EMS\CoreBundle\ContentTransformer\HtmlStylesRemover;
+
+class HtmlStylesRemoverTest extends WebTestCase
+{
+    public function testRemoveSimpleInline()
+    {
+        $input = <<<HTML
```
dameert: Isn't there a way to provide an array of inputs, and an array of outputs that is used by 1 testcase. (in console the number of tests will be the same, but you only have to program the flow once)

dameert: However, not sure if it will be possible to "customize" the error message for each case, which is now handled through the function names.

ContentTransformer/ContentTransformInterface.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace EMS\CoreBundle\ContentTransformer;
+
+interface ContentTransformInterface
+{
+    public function canTransform(ContentTransformContext $contentTransformContext): bool;
+    public function transform(string $input): string;
```
dameert: To be discussed, but I think every function should have the ContentTransformContext as parameter

ContentTransformer/HtmlStylesRemover.php
```diff
@@ -0,0 +1,118 @@
...
+
+    /** @var \DOMXPath */
+    protected $xpath;
+
+    public function __construct(string $classNamePrefix = 'removable-style-')
+    {
+        $this->classNamePrefix = $classNamePrefix;
+    }
+
+    public function canTransform(ContentTransformContext $contentTransformContext = null): bool
```
dameert: Why would we allow this to be NULL?

ContentTransformer/HtmlStylesRemover.php
```diff
@@ -0,0 +1,118 @@
...
+    {
+        if (!$contentTransformContext) {
+            return false;
+        }
+
+        foreach ($contentTransformContext->get() as $dataFieldType) {
```
dameert: to be refactored once the context object is redefined

dameert: Will review the rest of this file after the changes of Context are present

ContentTransformer/HtmlStylesRemover.php
```diff
@@ -23,33 +23,27 @@ public function __construct(string $classNamePrefix = 'removable-style-')
         $this->classNamePrefix = $classNamePrefix;
     }
 
-    public function canTransform(ContentTransformContext $contentTransformContext = null): bool
+    public function canTransform(ContentTransformContext $contentTransformContext): bool
     {
-        if (!$contentTransformContext) {
+        if (!$contentTransformContext->getDataFieldType() instanceof WysiwygFieldType) {
```
dameert: return $contentTransformContext->getDataFieldType() instanceof WysiwygFieldType;

ContentTransformer/HtmlStylesRemover.php
```diff
@@ -23,33 +23,27 @@ public function __construct(string $classNamePrefix = 'removable-style-')
...
-        }
+        return true;
     }
 
-    public function transform(string $input): string
+    public function transform(ContentTransformContext $contentTransformContext): string
     {
-        $this->input = $input;
-        $this->doc = $this->initDocument($input);
+        $this->input = $contentTransformContext->getData();
```
dameert: Why are we keeping state in this class? 
Can't we pass ->getData() as context to the other functions?

ContentTransformer/HtmlStylesRemover.php
```diff
@@ -23,33 +23,27 @@ public function __construct(string $classNamePrefix = 'removable-style-')
...
         $this->xpath = new \DOMXPath($this->doc);
 
         return $this->removeHtmlStyles();
     }
 
-    public function changed(string $output): bool
+    public function changed(ContentTransformContext $contentTransformContext): bool
     {
-        if ($this->input === $output) {
+        if ($this->input === $contentTransformContext->getTransformedData()) {
```
dameert: we can ask this on the context contentTransformContext->hasChanges() because it has $data and $transformedData as properties.

tests/ContentTransformer/HtmlStylesRemoverTest.php
dameert: We should discuss about this, I have yet another change in my head.

Service/TransformContentTypeService.php
```diff
@@ -69,9 +65,11 @@ public function transform(ContentType $contentType): \Generator
...
 
                     $transformer = $this->getTransformer($dataField);
-                    if (!empty($transformer) && $transformer->canTransform(new ContentTransformContext([$dataFieldType]))) {
-                        $dataTransformed = $transformer->transform($data);
-                        if ($transformer->changed($dataTransformed)) {
+                    $contentTransformContext = ContentTransformContext::fromDataFieldType(\get_class($dataFieldType), $data);
+                    if (!empty($transformer) && $transformer->canTransform($contentTransformContext)) {
+                        $dataTransformed = $transformer->transform($contentTransformContext);
+                        $contentTransformContext->setTransformedData($dataTransformed);
+                        if ($transformer->changed($contentTransformContext)) {
```
dameert: Maybe the transformer does not need this changed function. But only the contentTransformContext? ($context->hasChanges())

PR #280, opened 2020-02-19
Feature / Snapshot

coppee:


Service/EnvironmentService.php
```diff
@@ -32,24 +41,106 @@ class EnvironmentService
...
+        $environment->setAlias($this->container->getParameter('ems_core.instance_id') . $environment->getName());
+        $environment->setManaged(true);
+        $environment->setSnapshot($snapshot);
+
+        try {
+            $em = $this->doctrine->getManager();
+            $em->persist($environment);
+            $em->flush();
+        } catch (\Exception $e) {
+            dump($e->getMessage());
```
dameert: dump!

Service/EnvironmentService.php
```diff
@@ -32,24 +41,106 @@ class EnvironmentService
...
+        }
+
+        $indexName = $environment->getAlias() . AppController::getFormatedTimestamp();
+        $this->client->indices()->create([
+            'index' => $indexName,
+            'body' => $this->getIndexAnalysisConfiguration(),
+        ]);
+
+        $contentTypeService = $this->container->get('ems.service.contenttype');
```
dameert: Injection via Dependency Injection? container->get is deprecated!

Service/EnvironmentService.php
```diff
@@ -63,6 +154,27 @@ public function getEnvironmentNames(): array
         return array_keys($this->getEnvironments());
     }
 
+    public function getNotSnapshotEnvironments(): array
```
dameert: Ok because POC has priority, but should be refactored using the strategy from #281

Service/EnvironmentService.php
```diff
@@ -204,4 +316,11 @@ public function getAllInMyCircle()
             return count(array_intersect($user->getCircles(), $environment->getCircles())) >= 1;
         });
     }
+
+    public function clearCache()
```
dameert: Not needed (to be removed) when using #281 maybe we should add an @deprecated to this function immediatly and reference 281 in it?

dameert (changes requested):

Service/EnvironmentService.php
coppee: Yes but no : if we use D.I. there is a Circular reference error (ems.service.contenttype has ems.service.environement as dependency)

dameert: Interesting (this is not good! :D )

dameert (approved):

dameert (changes requested):
Travis build failed (this error is triggered due to changes in the EnvironmentService file !)

PR #282, opened 2020-02-20
Feature / Content-type translation fields

coppee:


Form/Form/ContentTypeType.php
```diff
@@ -121,6 +121,18 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                         'Descending' => 'desc',
                     ],
             ]);
+            $builder->add('translationField', ContentTypeFieldPickerType::class, [
+                'required' => false,
+                'firstLevelOnly' => true,
```
dameert: Should check if this should be false?

theus77: No it must be set to true, as we assume that the translation field can't be in a nested field

PR #285, opened 2020-02-20
Don't automatically extract data from file bigger than 2MB

theus77:
Avoid to automatically extract data from file bigger than 2MB


Resources/assets/edit-revision.js
```diff
@@ -336,7 +337,7 @@ function FileSelectHandler(e) {
 }
 
 //file data extractor
-function FileDataExtrator(container) {
+function FileDataExtrator(container, forced=false) {
```
dameert: function name should start with a non-capital letter

dameert: As discussed, this function allready existed. A refactor would be too risky

PR #287, opened 2020-02-23
filter Display Options

theus77:
Avoid not required datafieldtype option when  an option is removed

### DIFF
--- a/Entity/FieldType.php
+++ b/Entity/FieldType.php
@@ -7,6 +7,8 @@ use Doctrine\ORM\Mapping as ORM;
 use EMS\CoreBundle\Entity\ContentType;
 use EMS\CoreBundle\Entity\Helper\JsonClass;
 use EMS\CoreBundle\Entity\Helper\JsonDeserializer;
+use EMS\CoreBundle\Form\DataField\DataFieldType;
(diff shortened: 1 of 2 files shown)


Entity/FieldType.php
```diff
@@ -724,4 +726,21 @@ protected function deserializeProperty(string $name, $value)
...
+        $dataFieldType->configureOptions($optionsResolver);
+        $defineOptions = $optionsResolver->getDefinedOptions();
+        $defineOptions[] = 'label';
+
+        $filtered = array_filter(
+            $this->getDisplayOptions(),
+            function ($value) use ($defineOptions) {
+                return in_array($value, $defineOptions);
+            },
+            ARRAY_FILTER_USE_KEY
```
dameert: Using this filter option indicates you might need aan object to handle stuff. Cfr. https://matthiasnoback.nl/2020/01/rules-for-working-with-arrays-and-custom-collection-classes-in-php/ However I should look into your getDisplayOptions function first before judging!

dameert: I notice that this is saved as a Json array. Which we can keep. But getDisplayOptions should pass that complete option array to the constructor of a DisplayOptions object which knows what options are available and makes the code cleaner wherever displayOptions is used.

PR #292, opened 2020-02-25
Better asset extractor

theus77:
Adjust the request timeout on forced request + save the session to avoid to block the user


Service/RestClientService.php
```diff
@@ -12,11 +12,11 @@ class RestClientService
      * @param string $baseUrl
      * @return \GuzzleHttp\Client
      */
-    public function getClient($baseUrl = null)
+    public function getClient($baseUrl = null, int $timeout = 30)
```
coppee: Add return type Client

PR #296, opened 2020-02-26
Debug demo 20200226

theus77:


Controller/ContentManagement/DataController.php
```diff
@@ -1494,35 +1494,58 @@ public function editRevisionAction($revisionId, Request $request, LoggerInterfac
...
+    {
+        $dataService->hasCreateRights($contentType);
 
         try {
-            $revision = $dataService->newDocument($contentType, $revision->getOuuid(), $jsonContent);
+            $revision = $dataService->newDocument($contentType, null, $rawData);
 
             return $this->redirectToRoute('revision.edit', [
                 'revisionId' => $revision->getId()
             ]);
```
coppee: (My bad !) Pull out the "happy" redirect from the try and push her at the end of the method ?

Controller/ContentManagement/DataController.php
```diff
@@ -1494,35 +1494,58 @@ public function editRevisionAction($revisionId, Request $request, LoggerInterfac
...
-
-            return $this->redirectToRoute('data.root', [
-                'name' => $contentType->getName()
-            ]);
         }
+
+        return $this->redirectToRoute('data.root', [
+            'name' => $contentType->getName()
+        ]);
```
coppee: Put the "unhappy" redirect in the catch ?

Controller/ContentManagement/DataController.php
theus77: Or not as the $revision migth not exist outside the try

Controller/ContentManagement/DataController.php
theus77: Or not as the $revision migth not exist outside the try

PR #297, opened 2020-02-27
Feat / Check translation unicity

coppee:


Service/DataService.php
```diff
@@ -863,6 +863,11 @@ public function finalizeDraft(Revision $revision, FormInterface &$form = null, $
             $form = $builder->getForm();
         }
 
+        if ($this->checkIfTranslationExist($revision, $form)) {
+            throw new CantBeFinalizedException("Cannot finalized this revision because this translation already exist");
+        }
```
theus77: ....because a similar translation is already defined"

Service/DataService.php
```diff
@@ -966,6 +971,67 @@ public function finalizeDraft(Revision $revision, FormInterface &$form = null, $
...
+    {
+        $localeFieldName = $revision->getContentType()->getLocaleField();
+        $translationFieldName = $revision->getContentType()->getTranslationField();
+
+        if ($localeFieldName === null || $translationFieldName === null) {
+            return false;
+        }
+
+        if (!isset($form->get('data')->getData()[$localeFieldName]) || !isset($form->get('data')->getData()[$translationFieldName])) {
+            return false;
```
theus77: You must check the raw data not the field in the form (as those fields might be hidden in tabsFieldType or ContainerFieldType

Service/DataService.php
```diff
@@ -966,6 +971,67 @@ public function finalizeDraft(Revision $revision, FormInterface &$form = null, $
...
+                                ],
+                            ],
+                        ]
+                    ]
+                ]
+            ]
+        ]);
+
+        if ($result['hits']['total'] === 0) {
+            return false;
```
theus77: return $result['hits']['total'] !== 0

PR #300, opened 2020-02-28
Feature / Content transformer workflow refactoring

coppee:
Some changes after test in local for the HIPPOC project

### DIFF
--- a/Command/AlignCommand.php
+++ b/Command/AlignCommand.php
@@ -277,7 +277,8 @@ class AlignCommand extends Command
             return;
         }
 
-        $target = $this->environmentService->getAliasByName($targetName);
+        $this->environmentService->clearCache();
+        $target = $this->environmentService->getByName($targetName);
         if ($target === false) {
             $message = \sprintf('Target environment "%s" not found', $targetName);
             $this->setTargetArgument($input, $message);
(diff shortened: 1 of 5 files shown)


Command/AlignCommand.php
```diff
@@ -277,7 +277,8 @@ private function checkTarget(InputInterface $input)
             return;
         }
 
-        $target = $this->environmentService->getAliasByName($targetName);
+        $this->environmentService->clearCache();
```
theus77: Is not possible to update the cache on create a new environment (for a future refactoring)

ContentTransformer/HtmlStylesRemover.php
```diff
@@ -62,18 +65,33 @@ private function getInnerNode(\DOMNode $node): \DOMNode
...
     {
-        $doc = new \DOMDocument();
+        $doc = new \DOMDocument('1.0', 'UTF-8');
         $doc->preserveWhiteSpace = false;
         $doc->formatOutput = true;
 
-        $doc->loadHtml(
-            $this->addTemporaryWrapper($input),
+        @$doc->loadHtml( // Need @ to bypass the htmlParseEntityRef warnings/errors
+            \mb_convert_encoding($this->addTemporaryWrapper($input), 'HTML-ENTITIES', 'UTF-8'),
```
theus77: Which kind of error do you face? Is not not better to mb_emcode the  $input? Or to to do this convert in the addTemporaryWrapper?

coppee: @theus77  It's for avoid this kind of warning :  "Warning: DOMDocument::loadHTML(): htmlParseEntityRef: expecting ';' in Entity, line: 15". 

It's seems to be a knowing workaround used by Drupal and even Symfony (https://github.com/symfony/dom-crawler/blob/master/Crawler.php#L1128)

PR #302, opened 2020-03-01
Test driven dev

theus77:
I think (almost sure) that those test scripts are not relevant.


coppee:
Just for the record, there are two flows. This one is for removing styles. The other flow is the content remover and it's a custom content transformer for instructions (in elasticms gitlab repo)

coppee:
This logic is in the custom elasticms repo.

PR #305, opened 2020-03-07
getAll is deprecated and it returns false (which is no more ok for th…

theus77:
…e choices options null or array)


dameert:
We should activate the pr restrictions also to the develop branches! Otherwise it will be too big of a review when going to master (and too late)

PR #309, opened 2020-03-11
Reusable file uploader

theus77:
The FileType is currently only usable with the RevisionType. The idea of this PR is to make it available in every other core's form


dameert:

PR #310, opened 2020-03-11
Improve exception ems message

theus77:


dameert:

AND the Travis Build is failing --> fix that first

PR #311, opened 2020-03-11
Fix/revision id can be null

theus77:


dameert:

dameert:

PR #315, opened 2020-03-12
WIP: Add revision copy command

Davidmattei:


Service/DataService.php
```diff
@@ -851,12 +851,12 @@ public function buildForm(Revision $revision)
      * @throws Exception
      * @throws Throwable
      */
-    public function finalizeDraft(Revision $revision, FormInterface &$form = null, $username = null, $computeFields = true)
+    public function finalizeDraft(Revision $revision, FormInterface &$form = null, string $username = null, bool $computeFields = true)
```
dameert:
```suggestion
    public function finalizeDraft(Revision $revision, ?FormInterface &$form = null, ?string $username = null, bool $computeFields = true)
```

dameert: Or not allow null for users, (as you removed the check on line 859)

Davidmattei: Fixed

PR #319, opened 2020-03-15
New activate contenttypes command

theus77:


Command/ActivateContentTypeCommand.php
```diff
@@ -0,0 +1,147 @@
+<?php
+
```
Davidmattei:
```suggestion
declare(strict_types=1);
```
For new files we should start doing this

Command/ActivateContentTypeCommand.php
```diff
@@ -0,0 +1,147 @@
...
+        if (!$input->getOption(self::OPTION_ALL) && count($types) == 0) {
+            $this->chooseTypes($input, $output);
+        }
+
+        if ($input->getOption(self::OPTION_ALL)) {
+            $this->optionAll($input);
+        }
+        $this->deactivate = $input->getOption(self::DEACTIVATE);
+        $this->force = $input->getOption(self::FORCE);
```
Davidmattei:
```suggestion
```
We are not interacting with these options, so you can just all getOption in the execute, also remove the 2 private properties 'deactivate' and 'force' on this command class

Command/ActivateContentTypeCommand.php
```diff
@@ -0,0 +1,147 @@
...
+            )
+            ->addOption(
+                self::FORCE,
+                null,
+                InputOption::VALUE_NONE,
+                sprintf('Activate the contenttypes even if the mapping is not up to date (flagged as draft)')
+            );
+    }
+
+    public function execute(InputInterface $input, OutputInterface $output)
```
Davidmattei:
```suggestion
    protected function execute(InputInterface $input, OutputInterface $output)
```

theus77: execute as protected function

Command/ActivateContentTypeCommand.php
theus77: Agree for the force but the deactivate option : see 30930983116d0cff3ff17bcd77bdedb57115827d

Davidmattei: Oh okay I see, you need the deactivate in the chooseTypes and interact function. Okay I can live with that :)

PR #348, opened 2020-05-14
feat/ldap-user

dameert:
|Q              |A  |
|---------------|---|
|Bug fix?       |N|
|New feature?   |Y|	
|BC breaks?     |Y|	
|Deprecations?  |N|	
|Fixed tickets  ||	

Allows login via LDAP, and creates a user in DB on first login.
To use this, we had to migrate LocalUserBundle into CoreBundle (as Core should be able to create new users, without having a circular dependency on localuserbundle)


Controller/UserController.php
```diff
@@ -312,4 +411,27 @@ private function getExistingRoles()
...
+     */
+    private function userExist($user, $action, $form): bool
+    {
+        /** @var UserManagerInterface $userManager */
+        $userManager = $this->get('fos_user.user_manager');
+        $exists = array('email' => $userManager->findUserByEmail($user->getEmail()), 'username' => $userManager->findUserByUsername($user->getUsername()));
+        $messages = array('email' => 'User email already exist!', 'username' => 'Username already exist!');
+        foreach ($exists as $key => $value) {
+            if ($value instanceof User) {
+                if ($action == 'add' or ($action == 'edit' and $value->getId() != $user->getId())) {
```
coppee, line 426:
```suggestion
                if ($action === 'add' || ($action === 'edit' && $value->getId() !== $user->getId())) {
```

coppee: (OK, it's legacy)

Service/ObjectChoiceCacheService.php
```diff
@@ -6,7 +6,7 @@
 use Elasticsearch\Client;
 use EMS\CommonBundle\Helper\EmsFields;
 use EMS\CoreBundle\Form\Field\ObjectChoiceListItem;
-use EMS\LocalUserBundle\Entity\User;
+use EMS\CoreBundle\Entity\User;
```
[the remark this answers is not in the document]
dameert: Out of scope of this refactoring. I move the files of LocalUserBundle to Core, and update Uses (but indeed this should never have been LocalUserBundle as core did not depend on that bundle! --> This use statement is for a /** @var inside a function */
-> I wil make a PR for that (we keep this comment open)

Security/CoreLdapUserProvider.php
```diff
@@ -0,0 +1,84 @@
...
+        $refreshedUser = $this->userService->getUser($user->getUsername(), false);
+
+        if (!$refreshedUser instanceof SymfonyUserInterface) {
+            throw new UnsupportedUserException(sprintf('Instances of "%s" are not supported.', \get_class($refreshedUser)));
+        }
+
+        return $refreshedUser;
+    }
+
+    public function supportsClass($class)
```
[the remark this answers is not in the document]
dameert: I think I was not allowed to do that (I will retest this)

Service/UserService.php
```diff
@@ -101,12 +103,23 @@ public function getUser($username, $detachIt = true)
         return $user;
     }
     
-    public function getCurrentUser(bool $detach = true): User
+    public function getCurrentUser(bool $detach = true): UserInterface
     {
-        if (!$this->currentUser) {
-            $username = $this->tokenStorage->getToken()->getUsername();
-            $this->currentUser = $this->getUser($username, $detach);
+        if ($this->currentUser) {
```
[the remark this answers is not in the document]
dameert: Indeed, that would be better, but I think that refactoring this one would be quite substantial.

Security/Authenticator.php
```diff
@@ -0,0 +1,86 @@
...
+            'error' => ['Unauthorized Error'],
+        ]))->setStatusCode(401);
+        return $response;
+    }
+
+    public function successResponse(string $content): Response
+    {
+        $response = new Response();
+        $response->headers->set('Content-Type', 'application/json');
+        $response->setContent($content)->setStatusCode(401);
```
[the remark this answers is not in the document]
dameert: Good catch! fixed it

Security/CoreLdapUserProvider.php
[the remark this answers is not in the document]
dameert: Ok seems like it works, It's fixed as part of my next PR that I have prepared: 
https://github.com/dameert/EMSCoreBundle/pull/2

Service/ObjectChoiceCacheService.php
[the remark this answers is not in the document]
dameert: Fixed it in this one ;)

Security/CoreLdapUser.php
```diff
@@ -45,24 +45,33 @@ public function __toString(): string
...
+    public static function fromLdap(UserInterface $ldapUser, string $emailField): CoreLdapUser
     {
         if (! $ldapUser instanceof SymfonyLdapUser) {
             throw new \RuntimeException(\sprintf('Could not create ldap user. Instance should be of type %s', SymfonyLdapUser::class));
         }
 
         $user = new static();
-        $user->created = new \DateTime('now');
+        $user->circles = [];
+        $user->created = $user->modified = new \DateTime('now');
```
Davidmattei, line 56: ugly

Controller/UserController.php
```diff
@@ -9,20 +9,23 @@
...
 use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
 use Symfony\Component\Form\Extension\Core\Type\EmailType;
+use Symfony\Component\Form\Extension\Core\Type\PasswordType;
+use Symfony\Component\Form\Extension\Core\Type\RepeatedType;
+use Symfony\Component\Form\Extension\Core\Type\TextareaType;
 use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Routing\Annotation\Route;
+use EMS\LocalUserBundle\Entity\User;
```
Davidmattei: sort uses

Controller/UserController.php
```diff
@@ -37,6 +40,103 @@ public function indexAction()
         ]);
     }
 
+    /**
+     *
+     * @Route("/user/add", name="user.add")
+     */
+    public function addUserAction(Request $request)
```
Davidmattei, line 48: add return type: response?

Security/CoreLdapUser.php
dameert: Having them at 1 line, or using \DateTime ? :)

Controller/UserController.php
```diff
@@ -312,4 +412,27 @@ private function getExistingRoles()
     {
         return $this->getUserService()->getExistingRoles();
     }
+
+    /**
+     * Test if email or username exist return on add or edit Form
+     */
+    private function userExist ($user, $action, $form) {
```
Davidmattei: return type bool?

Controller/UserController.php
```diff
@@ -37,6 +40,103 @@ public function indexAction()
...
+            ->getForm();
+
+        $form->handleRequest($request);
+
+        if ($form->isSubmitted() && $form->isValid()) {
+            /** @var $userManager \FOS\UserBundle\Model\UserManagerInterface */
+            $userManager = $this->get('fos_user.user_manager');
+
+            $continue = TRUE;
+            $continue = $this->userExist($user, 'add', $form);
```
Davidmattei, line 121: userExist should allways return bool, make TRUE => true, FALSE => false

Form/UserProfileType.php
```diff
@@ -0,0 +1,85 @@
...
+			->add ('displayName')
+			->add ('emailNotification', CheckboxType::class, [
+					'required' => false,
+			])
+			->add ('layoutBoxed')
+			->add ('sidebarMini')
+			->add ('sidebarCollapse')
+			->remove('username');
+		
+// 		if($this->tokenStorage->getToken()->getUser()->getAllowedToConfigureWysiwyg()){
```
Davidmattei: ?? remove

Controller/UserController.php
dameert: true true, but this line is no longer in the UserController.php

Entity/User.php
```diff
@@ -0,0 +1,466 @@
+<?php
```
Davidmattei: declare strict

dameert: migrated as-is from local-user, strict typing will give issue on the $authTokens variable which I won't readd in this pr

Controller/UserController.php
dameert: Allready done in a later commit ;)

Controller/UserController.php
dameert: Was allready done, but I removed the line 121 now

Form/UserProfileType.php
dameert: done

Controller/UserController.php
dameert: Indeed, should be done. Migrated this function "as-is" the other functions in this file don't have return type either.

Security/CoreLdapUser.php
dameert: fixed

PR #360, opened 2020-05-24
Feat/update ct from json

theus77:
|Q              |A  |
|---------------|---|
|Bug fix?       |N|
|New feature?   |Y|	
|BC breaks?     |N|	
|Deprecations?  |N|	
|Fixed tickets  |N|	

New form/route to update a content type from a JSON file

### DIFF
--- a/Entity/Helper/JsonClass.php
+++ b/Entity/Helper/JsonClass.php
@@ -3,6 +3,7 @@
 namespace EMS\CoreBundle\Entity\Helper;
 
 use Doctrine\ORM\PersistentCollection;
+use EMS\CoreBundle\Entity\ContentType;
 
 class JsonClass implements \JsonSerializable
 {
@@ -84,10 +85,13 @@ class JsonClass implements \JsonSerializable
         ];
     }
 
-    public function jsonDeserialize()
+    public function jsonDeserialize(object $object = null)
     {
         $reflectionClass = new \ReflectionClass($this->class);
-        $instance = $reflectionClass->newInstance(...$this->constructorArguments);
+        $instance = $object;
+        if ($instance === null) {
+            $instance = $reflectionClass->newInstance(...$this->constructorArguments);
+        }
 
         foreach ($this->properties as $name => $value) {
             if (! $reflectionClass->hasProperty($name)) {
--- a/Service/ContentTypeService.php
+++ b/Service/ContentTypeService.php
@@ -15,9 +15,13 @@ use EMS\CoreBundle\Entity\SingleTypeIndex;
 use EMS\CoreBundle\Exception\ContentTypeAlreadyExistException;
 use EMS\CoreBundle\Form\DataField\DataFieldType;
 use EMS\CoreBundle\Repository\ContentTypeRepository;
+use EMS\CoreBundle\Repository\FieldTypeRepository;
 use EMS\CoreBundle\Repository\SingleTypeIndexRepository;
+use EMS\CoreBundle\Repository\TemplateRepository;
+use EMS\CoreBundle\Repository\ViewRepository;
 use Psr\Log\LoggerInterface;
 use Symfony\Component\Form\FormRegistryInterface;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 use Symfony\Component\Translation\TranslatorInterface;
 
 class ContentTypeService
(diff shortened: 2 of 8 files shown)


Entity/Helper/JsonClass.php
```diff
@@ -84,10 +85,13 @@ public function jsonSerialize()
         ];
     }
 
-    public function jsonDeserialize()
+    public function jsonDeserialize(object $object = null)
```
dameert, line 88: It is confusing that deserializing an object allows for another object to be it's deserialisation.

I Think you simply need another function that does the deserialisation of the properties when you allready have an object!

`foreach ($this->properties as $name => $value) {
            if (! $reflectionClass->hasProperty($name)) {
                continue;
            }
            $instance->deserialize($name, $value);
        }`

Service/ContentTypeService.php
```diff
@@ -426,10 +430,30 @@ public function getAllTypes()
...
+
+        $updatedContentType = $this->contentTypeFromJson($json, $environment, $contentType);
+        $this->importContentType($updatedContentType);
+    }
+
+    public function contentTypeFromJson(string $json, Environment $environment, ContentType $contentType = null): ContentType
     {
         $meta = JsonClass::fromJsonString($json);
-        $contentType = $meta->jsonDeserialize();
+        $contentType = $meta->jsonDeserialize($contentType);
```
dameert, line 456: As discussed in the JsonClass file, this should be a call to $meta->deserializeProperties($contentType). However this should not be called from here (as this function should only care to create contentTypeFromJson (factor out the code you need to a separate function?)

Service/ContentTypeService.php
```diff
@@ -426,10 +430,30 @@ public function getAllTypes()
         return implode(',', array_keys($this->contentTypeArrayByName));
     }
 
-    public function contentTypeFromJson(string $json, Environment $environment): ContentType
+
+    public function updateFromJson(ContentType $contentType, string $json, bool $isDeleteExitingTemplates, bool $isDeleteExitingViews): void
```
dameert, line 434: the analogy with contentTypeFromJson is distracting because it indicates the creation of a ContentType, from a Json input. updateFromJson indicates the creation of Update from Json input which is incorrect (maybe just "updateContentType"? As the function is updating the contentType passed as argument).


Also small fixes in naming of the parameters.

dameert, line 434:
```suggestion
    public function updateContentType(ContentType $contentType, string $json, bool $shoudDeleteExitingTemplates, bool $shouldDeleteExitingViews): void
```

Service/ContentTypeService.php
```diff
@@ -426,10 +430,30 @@ public function getAllTypes()
...
+    {
+        $this->deleteFields($contentType);
```
dameert, line 436: What happens if your program would be interrupted at this stage? Your content types fields would be removed, and the update never executed..

Service/ContentTypeService.php
```diff
@@ -438,13 +462,65 @@ public function contentTypeFromJson(string $json, Environment $environment): Con
...
     }
 
+    private function deleteFields(ContentType $contentType): void
+    {
+        $em = $this->doctrine->getManager();
+        $contentType->unsetFieldType();
+        /** @var FieldTypeRepository $fieldRepo */
+        $fieldRepo = $em->getRepository('EMSCoreBundle:FieldType');
+        $fields = $fieldRepo->findBy([
+            'contentType' => $contentType
```
dameert, line 472:
```suggestion
            'contentType' => $contentType->getName()
```

It's clearer for the programmer to be explicit on the value that is passed (in stead of relying on the __toString() convention). Also the IDE now knows that there is a dependency between this function and the name of the contentType

Service/ContentTypeService.php
```diff
@@ -426,10 +430,30 @@ public function getAllTypes()
...
+            $this->deleteViews($contentType);
+        }
+
+        $environment = $contentType->getEnvironment();
+        if (! $environment instanceof Environment) {
+            throw new NotFoundHttpException('Environment not found');
+        }
+
+        $updatedContentType = $this->contentTypeFromJson($json, $environment, $contentType);
+        $this->importContentType($updatedContentType);
```
dameert, line 450: The current strategy is to remove everything unused from the database. Then update from the json (or only the properties from the json). By refactoring (my other comments) this behaviour would be more clear. 

I have a question however. Why not create a ContentTypeFromJson with the new $json file. Then compare the currenct ContentType with the newly created one $oldContentType-->merge(Contenttype $ct) and then persist this updated representation of the contenttype in the database?

That would solve the inconsistent db state created at line 436.

PR #393, opened 2020-09-03
Fix: Doctrine AuthToken to User relation error

theus77:
|Q              |A  |
|---------------|---|
|Bug fix?       |Y|
|New feature?   |N|	
|BC breaks?     |N|	
|Deprecations?  |N|	
|Fixed tickets  |N|

### DIFF
--- a/Entity/AuthToken.php
+++ b/Entity/AuthToken.php
@@ -37,7 +37,7 @@ class AuthToken
     protected $modified;
 
     /**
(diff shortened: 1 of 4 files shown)


Davidmattei (changes requested):
In the constructor you should already define the collection: 

$this->authTokens = new ArrayCollection();

Then phpstan will also be happy :)

Davidmattei (approved):

PR #404, opened 2020-09-27
feat: submissions overview mail

laurensmertens:
|Q              |A  |
|---------------|---|
|Bug fix?       |No|
|New feature?   |Yes|	
|BC breaks?     |No|	
|Deprecations?  |No|	
|Fixed tickets  |No|	

Added a service method to send a list of submissions by mail and a command to use it.


Command/EmailSubmissionsCommand.php
```diff
@@ -0,0 +1,57 @@
...
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $formInstance = \strval($input->getArgument('formInstance'));
+        $templateId = \strval($input->getArgument('templateId'));
+        $emails = (array) $input->getArgument('emails');
+
+        $submissions = $this->formSubmissionService->getFormInstanceSubmissions($formInstance);
+
+        $this->formSubmissionService->mailSubmissions($submissions, $formInstance, $templateId, $emails);
```
dameert, line 50: I think that a formSubmissionService should fetch the data of all submissions. And a separate MailService should send the mail. That way sending mails becomes reusable without extra effort.

And we could imagine to export formSubmissions via other channels then mails.

Service/Form/Submission/FormSubmissionService.php
```diff
@@ -91,4 +131,52 @@ public function submit(FormSubmissionRequest $submitRequest): array
...
+            return 'There are no submissions for this form';
+        }
+
+        try {
+            $template = $this->templateService->init($templateId)->getTemplate()->getBody();
+        } catch (\Exception $e) {
+            $template = "Error in body template: " . $e->getMessage();
+        }
+
+        $rows = '<tr>
```
dameert, line 152: html should be generated using twig files

dameert (changes requested):

Resources/views/email/submissions.email.twig
```diff
@@ -0,0 +1,111 @@
...
+                    </td>
+                </tr>
+                <tr>
+                    <td class="email-intro"
+                        style="color: #3a383a; font-family: Georgia, serif; font-size: 18px; line-height: 27px; padding-top: 5px;">
+                        <table border="1">
+                            <tr>
+                                <th>Label</th>
+                                <th>Submission Date</th>
+                                <th>Expiry Date</th>
```
dameert, line 86: With this information we can not 100% map a consentment to a user if he would ask for it.
The uuid of the submission is more important than the label. I think we should have both here.

Resources/views/email/submissions.email.twig
```diff
@@ -0,0 +1,111 @@
...
+                            {% for submission in submissions %}
+                                <tr>
+                                    <td>{{ submission.label }}</td>
+                                    <td>{{ submission.created|date('d/m/Y') }}</td>
+                                    <td>{{ submission.expireDate|date('d/m/Y') }}</td>
+                                    {% for key, value in submission.data %}
+                                        {% if value  is not iterable %}
+                                            <td>{{ value }}</td>
+                                        {% endif %}
+                                    {% endfor %}
```
dameert, line 100: What value would that be?

Resources/views/email/submissions.email.twig
```diff
@@ -0,0 +1,111 @@
...
+                    <td class="email-intro"
+                        style="color: #3a383a; font-family: Georgia, serif; font-size: 18px; line-height: 27px; padding-top: 5px;">
+                        <table border="1">
+                            <tr>
+                                <th>Label</th>
+                                <th>Submission Date</th>
+                                <th>Expiry Date</th>
+                                {% for key, value in submissions.0.data %}
+                                    <th> {{ key }}</th>
+                                {% endfor %}
```
dameert, line 89: What are you exposing this way?

dameert (changes requested):

Resources/views/email/submissions.email.twig
theus77: All data in the submission I'm afraid. And there is a bug if submissions does not have the same fields in the same order. I understand that we need to get some information in order to know who gave its approval. Maybe we should parsed all fields of all submissions to get a list of form fields having one of those patterns in their name:['name', 'phone', 'email']?

Resources/views/email/submissions.email.twig
laurensmertens: The title of the column, for example "firstname" or "email", but as discussed will be removed.

Resources/views/email/submissions.email.twig
laurensmertens: Good point, at first I had it per form, but then it was indeed an issue.
Shouldn't be a problem anymore if ID, label & dates are enough.

theus77 (approved):

Command/EmailSubmissionsCommand.php
```diff
@@ -0,0 +1,74 @@
...
+                null,
+                InputOption::VALUE_OPTIONAL,
+                'From which form instance do you want to mail submissions? Defaults to all.'
+            );
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $emails = (array) $input->getArgument('emails');
+        $formInstance = \strval($input->getOption('formInstance'));
```
dameert, line 57:
```suggestion
        $formInstance = (string) $input->getOption('formInstance');
```

Repository/FormSubmissionRepository.php
```diff
@@ -35,6 +35,35 @@ public function findAllUnprocessed(): array
...
+    public function findAllFormSubmissions(): array
+    {
+        $qb = $this->createQueryBuilder('fs');
+        $qb
+            ->andWhere($qb->expr()->isNotNull('fs.data'))
+            ->orderBy('fs.created', 'desc');
+
+        return $qb->getQuery()->getArrayResult();
+    }
+
```
dameert, line 66: findFormSubmissions(?string $formInstance = null)

Service/Form/Submission/FormSubmissionService.php
```diff
@@ -91,4 +123,23 @@ public function submit(FormSubmissionRequest $submitRequest): array
 
         return ['submission_id' => $formSubmission->getId()];
     }
+
+    /**
+     * @param array<FormSubmission> $submissions
+     * @return string
```
dameert, line 129:
```suggestion
```

Service/Form/Submission/FormSubmissionService.php
```diff
@@ -91,4 +123,23 @@ public function submit(FormSubmissionRequest $submitRequest): array
...
+            return 'There are no submissions for this form';
+        }
+
+        try {
+            $template = $this->twig->render('@EMSCore/email/submissions.email.twig', ['submissions' => $submissions]);
+        } catch (\Exception $e) {
+            $template = $this->twig->createTemplate("Error in body template: " . $e->getMessage())->render();
+        }
+
+        return $template;
```
dameert, line 143:
```suggestion
        try {
           return $this->twig->render('@EMSCore/email/submissions.email.twig', ['submissions' => $submissions]);
        } catch (\Exception $e) {
            return $this->twig->createTemplate("Error in body template: " . $e->getMessage())->render();
        }
```

Service/MailerService.php
```diff
@@ -0,0 +1,36 @@
...
+
+namespace EMS\CoreBundle\Service;
+
+use Swift_Mailer;
+use Swift_Message;
+
+class MailerService
+{
+    const EMAIL_FROM = 'reporting@elasticms.test';
+    const NAME_FROM = 'ElasticMS';
```
dameert, line 11: Can't be hard coded. Load this from the configuration (we already have email information in the .env)

dameert (changes requested):

Service/MailerService.php
dameert: see:
EMS_FROM_EMAIL_ADDRESS='ems@test'
EMS_FROM_EMAIL_NAME='EMS'

Command/EmailSubmissionsCommand.php
laurensmertens: PHPStan does not allow this - Cannot cast array<string>|bool|string|null to string.

Davidmattei (approved):

PR #423, opened 2020-10-26
feat: core versioning

Davidmattei:
|Q              |A  |
|---------------|---|
|Bug fix?       |no|
|New feature?   |yes|	
|BC breaks?     |no|	
|Deprecations?  |no|	
|Fixed tickets  |no|	

**Versioning Data structure**
- [x] eaf6b07 | New textarea field in "content type" entity named "version tags" 
- [x] 6bff878 | New text field in "content type" entity named "version date from field"
- [x] 6bff878 | New text field in "content type" entity named "version date to field"
- [x] 03ada27 | Reserved mapping field "_version_tag" => keyword and "_version_uuid" => keyword
- [x] c488bd8 | New version_uuid field into the "revision" entity

**Versioning steps**
- [x] 0d5953e | On finalization of a document and if there is more than one version tag, show popup and select the tag 
- [x] 0d5953e | On finalization of an existing document and if there is version tag defined, show popup and select one of the tag or silent publication. A silent publication is just a "finalization" (publish into the default environment - without doing anything)
- [x] c488bd8 | On finalization, set the "_version_tag" and the "_version_uuid"  
- [x] c488bd8 | On finalization, if a version tag is selected by the user :

> - Timestamp now() is defined
> - Initialize and finalize a new document from the current draft revision
> - _version_tag selected by the user
> - "version date from field" = Timestamp now()
> - Current draft is discarded
> - New draft is initialized and finalized with "version date to field" = Timestamp now()

TODO:
- [X] 272b6b2 | Doctrine Migrations
- [X] 4b8e1c3 | Build assets
- [x] dbe12ed | Add documentation


Entity/ContentType.php
```diff
@@ -420,6 +413,27 @@ class ContentType extends JsonDeserializer implements \JsonSerializable
      */
     protected $createLinkDisplayRole = 'ROLE_USER';
 
+    /**
+     * @var string[]
+     *
+     * @ORM\Column(name="version_tags", type="json_array", nullable=true)
+     */
+    protected $versionTags = [];
```
theus77, line 421: Why multiple version tags?

Entity/ContentType.php
```diff
@@ -172,13 +172,6 @@ class ContentType extends JsonDeserializer implements \JsonSerializable
      */
     protected $colorField;
 
-    /**
-     * @var string
-     *
-     * @ORM\Column(name="parentField", type="string", length=100, nullable=true)
-     */
-    protected $parentField;
```
theus77, line 180: Does this not breaking the relation between ContentType and FieldType? No Doctrine warning?

Entity/Revision.php
```diff
@@ -770,6 +782,15 @@ public function getContentType()
         return $this->contentType;
     }
 
+    public function getContentTypeName(): string
+    {
+        if (null === $this->contentType) {
+            throw new \Exception('No contentType for revision!');
```
theus77, line 788: What'sbetter between \Exception and \RuntimeException in this case?

Resources/assets/edit-revision.js
```diff
@@ -226,6 +226,13 @@ function addEventListeners(target){
 
     new EmsListeners(target.get(0), onFormChange);
 
+    target.find('button#btn-publish-version').on('click', function(e) {
+        e.preventDefault();
+        $('#publish-version-modal').modal('show');
+
+        console.debug("cool");
```
theus77, line 233: To remove

Resources/views/contenttype/edit.html.twig
```diff
@@ -132,4 +156,20 @@
...
+            a2lix_lib.sfCollection.init({
+                collectionsSelector: '#content_type_versionTags',
+                manageRemoveEntry: true,
+                lang: {
+                    add: 'Add version',
+                    remove: 'X'
+                }
+            });
+        });
+    </script>
```
theus77, line 174: Move to a specific js app?

Entity/ContentType.php
Davidmattei: Its on the contentType: Major, Minor, ..
On the publish you can only select Silent or ONE of the version tags

Entity/ContentType.php
Davidmattei: Will check it again ...

Resources/views/contenttype/edit.html.twig
Davidmattei: 2fe4fb7

Resources/assets/edit-revision.js
Davidmattei: 2fe4fb7

Entity/Revision.php
Davidmattei: 2fe4fb7

theus77 (approved):

Entity/Revision.php
```diff
@@ -1078,4 +1099,54 @@ public function hasVersionTags() : bool
...
+    {
+        return $this->rawData[Mapping::VERSION_TAG] ?? null;
+    }
+
+    public function getVersionTagDefault(): string
+    {
+        $versionTags = $this->contentType ? $this->contentType->getVersionTags() : [];
+
+        if (!isset($versionTags[0]) || !is_string($versionTags[0])) {
+            throw new \Exception(sprintf('No version tags found for contentType %s', $this->getContentTypeName()));
```
dameert, line 1118: error should say "call hasVersionTag first"

Davidmattei: Fixed

Form/DataField/DataFieldType.php
```diff
@@ -183,13 +183,17 @@ public function getDefaultOptions($name)
...
+        /** @var FieldType $fieldType */
+        $fieldType = $options['metadata'];
+
+        if (\in_array($fieldType->getName(), $options['disabled_fields'] ?? [], true)) {
+            return true;
+        }
+
         if (strcmp('cli', php_sapi_name()) === 0) {
             return false;
         }
```
theus77, line 196: I would not disable it in CLI mode (perf and no meaning)
```suggestion
        if (strcmp('cli', php_sapi_name()) === 0) {
            return false;
        }

        /** @var FieldType $fieldType */
        $fieldType = $options['metadata'];

        if (\in_array($fieldType->getName(), $options['disabled_fields'] ?? [], true)) {
            return true;
        }

```

Davidmattei: fix in #344caf

Command/RecomputeCommand.php
```diff
@@ -125,15 +125,15 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         }
 
         if (!$input->getOption('continue') || $input->getOption('cron')) {
-            $forceFlag = $input->getOption('force');
+            $forceFlag = (bool) $input->getOption('force');
```
theus77, line 128: Is it not better like this?
```suggestion
            $forceFlag = $input->getOption('force') === true;
```

theus77 (approved):

PR #424, opened 2020-10-26
feat: Phpcs fos

AdrienHq:
|Q              |A  |
|---------------|---|
|Bug fix?       |N|
|New feature?   |Y|	
|BC breaks?     |N|	
|Deprecations?  |N|	
|Fixed tickets  |N|	

Activer le native_function_invocation


.travis.yml
```diff
@@ -32,6 +32,6 @@ script:
   # Search for PHP syntax errors.
   - find -L . -path ./vendor -prune -o -name '*.php' -print0 | xargs -0 -n 1 -P 4 php -l
   # Run PHPCS
-  - if [[ "$SNIFF" == "1" ]]; then vendor/bin/phpcs . --standard=./phpcs.xml.dist --extensions=php,lib,inc; fi
+  - if [[ "$SNIFF" == "1" ]]; then vendor/bin/php-cs-fixer fix; fi
```
dameert, line 35:
```suggestion
  - if [[ "$SNIFF" == "1" ]]; then vendor/bin/php-cs-fixer fix --dry-run --diff; fi
```

cfr https://github.com/ems-project/EMSSubmissionBundle/blob/412817da1f4962ea93018af27622637706a714e3/.travis.yml#L50

Travis can only check if PHPCS is respected, it should not try to fix issues in this stage.
However you did not run the phpcs instruction as the rules did not get applied to the code!

dameert (changes requested):

.php_cs.dist
```diff
@@ -0,0 +1,24 @@
...
+    ->in(__DIR__)
+    ->exclude('vendor')
+    ->exclude('tests/tmp')
+;
+
+return PhpCsFixer\Config::create()
+    ->setRules([
+        '@Symfony' => true,
+        'declare_strict_types' => true,
+        'final_class' => true,
```
dameert, line 19:
```suggestion
       // 'final_class' => true,
```

Because this rule needs a review for every class that is used as base for others (extends)

dameert (changes requested):
Do not forget to launch phpcs locally, to apply the rules (otherwise travis won't be happy)

PR #452, opened 2020-11-09
Rector

dameert:
|Q              |A  |
|---------------|---|
|Bug fix?       |N|
|New feature?   |N|	
|BC breaks?     |N|	
|Deprecations?  |N|	
|Fixed tickets  |cleaning up dead code with rector #no-ticket|	

Configuration for rector. Applied with dead code removal on the /src/Command directory.

TODO

- [x] adapt services.yml for cleaned Command constructors
- [x] double check that removed Command constants are not used by our other bundles


src/Controller/Views/CalendarController.php
```diff
@@ -17,7 +17,7 @@ class CalendarController extends AppController
      * @param Request $request
      * @return Response
      *
-     * @Route("/views/calendar/replan/{view}.json", name="views.calendar.replan", defaults={"_format": "json"}, methods={"POST"})
+     * @Route("/views/calendar/replan/{view}.json", name="views.calendar.replan", methods={"POST"})
```
theus77, line 20: _format forces the response mime type I would not remove it without testing it

src/Controller/Views/CalendarController.php
```diff
@@ -73,7 +73,7 @@ public function updateAction(View $view, Request $request)
      * @return Response
      * @throws \Exception
      *
-     * @Route("/views/calendar/search/{view}.json", name="views.calendar.search", defaults={"_format": "json"}, methods={"GET"})
+     * @Route("/views/calendar/search/{view}.json", name="views.calendar.search", methods={"GET"})
```
theus77, line 76: _format forces the response mime type I would not remove it without testing it

theus77 (commented):
I pretty sure that the _format sets the mime type automagically. You should set the mime type manually before returning the response bag.

src/Controller/Views/CalendarController.php
dameert: https://symfony.com/doc/current/routing.html#special-parameters

Indeed I will roll back these _format removals (did them by hand on only two places I think).

dameert: Oh yeah I removed them because it was hard coded in the route itself.
Changing to a minimal diff in my next commit.

However I think the proper solution would be this: 
```suggestion
     * @Route("/views/calendar/replan/{view}.{_format}", name="views.calendar.replan", requirements={"_format"="json"}, methods={"POST"})
```

dameert:
Ready to merge. Other folders will be checked and committed on another branch.

theus77 (approved):

PR #454, opened 2020-11-10
fix: make sure password values are available

laurensmertens:
|Q              |A  |
|---------------|---|
|Bug fix?       |Yes|
|New feature?   |No|	
|BC breaks?     |No|	
|Deprecations?  |No|	
|Fixed tickets  |No|	

Added a Credentials DTO to make sure the password values are available.


dameert (approved):
Accepting this as quick fix (currently not able to edit user page because it requires to always change password even if change-password fields are not present).

PR #460, opened 2020-11-16
refactor: cleaning of old storage service variables

theus77:
|Q              |A  |
|---------------|---|
|Bug fix?       |N|
|New feature?   |N|	
|BC breaks?     |Y|	
|Deprecations?  |Y|	
|Fixed tickets  |N|


laurensmertens (dismissed):

composer.json
```diff
@@ -61,7 +61,7 @@
 		"phpunit/phpunit" : "^7.1",
 		"symfony/test-pack": "^1.0",
 		"squizlabs/php_codesniffer": "3.4",
-		"rector/rector": "dev-master"
+		"rector/rector": "^0.8"
```
dameert, line 64: i fixed this to 0.8.56. As this is a dev tool I prefer it to not auto update on each build.

dameert: (already on develop branch)

src/Service/FileService.php
```diff
@@ -7,91 +7,34 @@
...
-
-    public function getStorageService(string $storageServiceId): ?StorageInterface
-    {
-        foreach ($this->storageManager->getAdapters() as $storageService) {
-            if ($storageService->__toString() == $storageServiceId) {
-                return $storageService;
-            }
-        };
-        return null;
+        $this->uploadMinimumNumberOfReplications = 10;
```
dameert, line 37: This is not new code, but should we not have this in config?
```suggestion
        $this->uploadMinimumNumberOfReplications = 10;
```

dameert (approved):
Toto update branch with develop and accept the composer.json rector version of develop.

src/Service/FileService.php
theus77: I thinking of adding a `read-only` and a `to-skip` flag in the EMS_STORAGE and implement a isReadOnly and  isToSkip methods in the storage services?

PR #465, opened 2020-11-18
WIP Ref/hidden storage adapters

theus77:
|Q              |A  |
|---------------|---|
|Bug fix?       |N|
|New feature?   |N|	
|BC breaks?     |N|	
|Deprecations?  |Y|	
|Fixed tickets  |N|


src/Command/CleanAssetCommand.php
```diff
@@ -66,13 +66,8 @@ protected function execute(InputInterface $input, OutputInterface $output): int
             foreach ($hashes as $hash) {
                 $usedCounter = $revRepo->hashReferenced($hash['hash']);
                 if ($usedCounter === 0) {
-                    /**@var StorageInterface $storage */
-                    foreach ($this->fileService->getStorages() as $storage) {
-                        $storage->remove($hash['hash']);
-                    }
```
Davidmattei, line 72: The remove is not necessary anymore?

Davidmattei (changes requested):

src/Command/CleanAssetCommand.php
theus77: :-) Glad you ask. 

If a file is referenced by 2 ems sharing the same storage you will delete it for both. If you want to clean you storage you should create another storage and ems:asset: synch it (and drop the old storage). As only referenced file will be synched! All the rest will be cleaned. I will add a note in the doc about this.

Davidmattei: Okay perfect! 👍

Davidmattei (approved):

PR #472, opened 2020-11-22
Ref/elastica migration part2

theus77:
|Q              |A  |
|---------------|---|
|Bug fix?       |N|
|New feature?   |Y|	
|BC breaks?     |N|	
|Deprecations?  |Y|	
|Fixed tickets  |N|	

44 classes with Elasticsearch\Client to go


src/Service/ContentTypeService.php
```diff
@@ -4,9 +4,13 @@
 
 use Doctrine\Bundle\DoctrineBundle\Registry;
 use Doctrine\ORM\NoResultException;
+use Elastica\Aggregation\Terms;
```
Davidmattei, line 7: Don't like the use of the Elastica namespace in the coreBundle.

We have a search object in CommonBundle add a method, addTerms ?

src/Service/DataService.php
```diff
@@ -12,12 +12,15 @@
 use Doctrine\ORM\OptimisticLockException;
 use Doctrine\ORM\ORMException;
 use Doctrine\ORM\QueryBuilder;
+use Elastica\Query\Term;
```
Davidmattei, line 15: Elastica in the coreBundle

Davidmattei (changes requested):
Import off the Elastica\Terms I would not do it, for the rest is a good PR!

src/Service/ContentTypeService.php
theus77: Fixed

src/Service/DataService.php
theus77: Fixed

Davidmattei (approved):

PR #485, opened 2020-11-27
ref: migrate DataController.php

theus77:
|Q              |A  |
|---------------|---|
|Bug fix?       |N|
|New feature?   |N|	
|BC breaks?     |N|	
|Deprecations?  |Y|	
|Fixed tickets  |N|


dameert:
It seems that the Semantic PR check requires "refactor and not ref" as prefix. We could accept this or add a customisation file to the code to allow "ref: "?

Davidmattei (approved):

PR #487, opened 2020-11-30
Ref/elastica migration part3

theus77:
|Q              |A  |
|---------------|---|
|Bug fix?       |Y|
|New feature?   |Y|	
|BC breaks?     |N|	
|Deprecations?  |Y|	
|Fixed tickets  |N|

### DIFF
--- a/src/Elasticsearch/Bulker.php
+++ b/src/Elasticsearch/Bulker.php
@@ -107,11 +107,11 @@ class Bulker
         return $this->send();
     }
 
-    public function indexDocument(DocumentInterface $document, string $index): bool
+    public function indexDocument(DocumentInterface $document, string $index, bool $upsert = false): bool
     {
         $body = $document->getSource();
 
-        return $this->index($document->getContentType(), $document->getId(), $index, $body);
+        return $this->index($document->getContentType(), $document->getId(), $index, $body, $upsert);
     }
 
     /**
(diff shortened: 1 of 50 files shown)


src/Elasticsearch/Bulker.php
```diff
@@ -96,53 +71,47 @@ public function setSize(int $size): Bulker
...
+            $action->setSource($body);
         }
-
+        $this->bulk->addAction($action);
         ++$this->counter;
 
         return $this->send();
     }
 
-    public function indexDocument(DocumentInterface $document, string $index, bool $upsert = false): bool
```
Davidmattei, line 132: upsert removed, please put it back

src/Twig/AppExtension.php
```diff
@@ -5,10 +5,12 @@
 use Caxy\HtmlDiff\HtmlDiff;
 use DateTime;
 use Doctrine\Bundle\DoctrineBundle\Registry;
-use Elasticsearch\Client;
+use Elastica\ResultSet;
```
Davidmattei, line 8: Don't we need to use the common result set? 

https://github.com/ems-project/EMSCommonBundle/blob/develop/src/Elasticsearch/Elastica/ResultSet.php

Davidmattei (dismissed):

src/Twig/AppExtension.php
theus77: It's an hook to make the client 5 working with es7

src/Elasticsearch/Bulker.php
theus77: Done

src/Twig/AppExtension.php
Davidmattei: okay aggree

Davidmattei (approved):

PR #575, opened 2021-02-06
chore: move doc

theus77:
|Q              |A  |
|---------------|---|
|Bug fix?       |N|
|New feature?   |N|	
|BC breaks?     |N|	
|Deprecations?  |N|	
|Fixed tickets  |N|	

I fell like it's better to move the bundle doc to the bundle's root  (just like other bundles)


Davidmattei (approved):
Yes we should this for all our bundles

PR #582, opened 2021-02-10
fix: nested copy_to _all field mapping

dameert:
|Q              |A  |
|---------------|---|
|Bug fix?       |Y|
|New feature?   ||	
|BC breaks?     ||	
|Deprecations?  ||	
|Fixed tickets  ||	

Nested fields should also be copied to the _all field

### DIFF
--- a/src/Service/Mapping.php
+++ b/src/Service/Mapping.php
@@ -267,11 +267,11 @@ class Mapping
     private function addCopyToAllField(array &$properties): void
     {
         foreach ($properties as $name => &$options) {
-            if (\in_array($options['type'] ?? null, ['text', 'keyword'], true)) {
+            if (\in_array(($options['type'] ?? null), ['text', 'keyword'], true)) {
                 $options['copy_to'] = \array_unique(\array_merge(['_all'], $options['copy_to'] ?? []));
                 continue;
             }
-            if (\in_array($options['type'] ?? null, ['nested'], true) && isset($options['properties'])) {
+            if (isset($options['properties'])) {
                 $this->addCopyToAllField($options['properties']);
             }
         }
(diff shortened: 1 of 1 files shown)


src/Service/Mapping.php
```diff
@@ -266,4 +262,18 @@ public function updateMapping(string $name, array $mappings, string $type): void
...
+    /** @param array<string, array<string, mixed>> $properties */
+    private function addCopyToAllField(array &$properties): void
+    {
+        foreach ($properties as $name => &$options) {
+            if (\in_array($options['type'] ?? null, ['text', 'keyword'], true)) {
+                $options['copy_to'] = \array_unique(\array_merge(['_all'], $options['copy_to'] ?? []));
+                continue;
+            }
+            if (\in_array($options['type'] ?? null, ['nested'], true)) {
+                $this->addCopyToAllField($options['properties']);
```
theus77, line 275:
```suggestion
                $this->addCopyToAllField($options['properties'] ?? []);
```

src/Service/Mapping.php
```diff
@@ -266,4 +262,18 @@ public function updateMapping(string $name, array $mappings, string $type): void
...
+
+    /** @param array<string, array<string, mixed>> $properties */
+    private function addCopyToAllField(array &$properties): void
+    {
+        foreach ($properties as $name => &$options) {
+            if (\in_array($options['type'] ?? null, ['text', 'keyword'], true)) {
+                $options['copy_to'] = \array_unique(\array_merge(['_all'], $options['copy_to'] ?? []));
+                continue;
+            }
+            if (\in_array($options['type'] ?? null, ['nested'], true)) {
```
Davidmattei, line 274:
```suggestion
            if (\in_array(($options['type'] ?? null), ['nested'], true)) {
```

Davidmattei (changes requested):

src/Service/Mapping.php
dameert: Same structure as the if test above, easier to extend. Discussed this also with Mathieu

Davidmattei (dismissed):

PR #586, opened 2021-02-11
Fix: post processing json menu nested

Davidmattei:
|Q              |A  |
|---------------|---|
|Bug fix?       |y|
|New feature?   |y|	
|BC breaks?     |n|	
|Deprecations?  |n|	
|Fixed tickets  |n|	

- Isolate post processing in new service
- Add giveContentType on Revision for not doing nullable checks on something that should not be null.
- If the nested container has a field called label, don't add a label.

### DIFF
--- a/src/Form/Form/RevisionJsonMenuNestedType.php
+++ b/src/Form/Form/RevisionJsonMenuNestedType.php
@@ -4,15 +4,15 @@ declare(strict_types=1);
 
 namespace EMS\CoreBundle\Form\Form;
 
+use EMS\CoreBundle\Entity\ContentType;
 use EMS\CoreBundle\Entity\FieldType;
+use EMS\CoreBundle\Form\DataField\TextStringFieldType;
 use EMS\CoreBundle\Form\DataTransformer\DataFieldModelTransformer;
 use EMS\CoreBundle\Form\DataTransformer\DataFieldViewTransformer;
 use Symfony\Component\Form\AbstractType;
-use Symfony\Component\Form\Extension\Core\Type\TextType;
 use Symfony\Component\Form\FormBuilderInterface;
 use Symfony\Component\Form\FormRegistryInterface;
 use Symfony\Component\OptionsResolver\OptionsResolver;
-use Symfony\Component\Validator\Constraints\NotBlank;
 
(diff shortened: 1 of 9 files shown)


src/Form/Form/RevisionJsonMenuNestedType.php
```diff
@@ -32,40 +32,39 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
     {
         /** @var FieldType $fieldType */
         $fieldType = $options['field_type'];
+        /** @var ContentType $contentType */
+        $contentType = $options['content_type'];
```
theus77, line 37:
```suggestion
        $contentType = $options['content_type'] ?? null;
        
        if (!$contentType instanceof ContentType) {
            throw new \RuntimeException('Unexpected ContentType object');
        }

```

src/Form/Form/RevisionJsonMenuNestedType.php
```diff
@@ -32,40 +32,39 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
...
-
-            $required = $child->getRestrictionOptions()['mandatory'] ?? false;
+        $labelChild = $fieldType->getChildren()->filter(fn (FieldType $c) => 'label' === $c->getName());
+        if (0 === $labelChild->count()) {
+            $labelFieldType = new FieldType();
+            $labelFieldType->setName('label');
+            $labelFieldType->setType(TextStringFieldType::class);
+            $labelFieldType->setOptions([
+                'displayOptions' => ['label' => 'Label'],
+                'restrictionOptions' => ['mandatory' => true],
```
theus77, line 45: I think that the label doesn't have to be mandatory

src/Form/Form/RevisionJsonMenuNestedType.php
Davidmattei: The option resolver has a allowType set on ContentType, so it will always be a ContentType instance

src/Form/Form/RevisionJsonMenuNestedType.php
Davidmattei: 52de5b1

theus77 (approved):

theus77 (approved):

PR #588, opened 2021-02-12
Feat/notification bulk

Davidmattei:
|Q              |A  |
|---------------|---|
|Bug fix?       |n|
|New feature?   |y|	
|BC breaks?     |n|	
|Deprecations?  |n|	
|Fixed tickets  |n|

### DIFF
--- a/src/Command/Notification/BulkActionCommand.php
+++ b/src/Command/Notification/BulkActionCommand.php
@@ -1,0 +1,147 @@
+<?php
+
+declare(strict_types=1);
+
+namespace EMS\CoreBundle\Command\Notification;
+
+use EMS\CommonBundle\Elasticsearch\Document\Document;
+use EMS\CommonBundle\Elasticsearch\Document\DocumentInterface;
+use EMS\CommonBundle\Service\ElasticaService;
+use EMS\CoreBundle\Entity\Environment;
+use EMS\CoreBundle\Service\EnvironmentService;
+use EMS\CoreBundle\Service\NotificationService;
+use EMS\CoreBundle\Service\Revision\RevisionService;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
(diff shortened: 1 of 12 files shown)


src/Command/Notification/BulkActionCommand.php
```diff
@@ -0,0 +1,147 @@
...
+
+        return 0;
+    }
+
+    /**
+     * @param array<mixed> $query
+     *
+     * @return DocumentInterface[]
+     */
+    private function searchDocuments(Environment $environment, array $query): array
```
theus77, line 126: Why nott returning an iterator? You load all document in memory

src/Entity/ContentType.php
```diff
@@ -1149,6 +1149,15 @@ public function getEnvironment()
...
 
+    public function giveEnvironment(): Environment
+    {
+        if (null === $this->environment) {
+            throw new \RuntimeException('Environment not found!');
+        }
+
+        return $this->environment;
+    }
+
```
theus77, line 1160: I love it

src/Entity/Template.php
```diff
@@ -708,12 +708,7 @@ public function setContentType(ContentType $contentType = null)
         return $this;
     }
 
-    /**
-     * Get contentType.
-     *
-     * @return \EMS\CoreBundle\Entity\ContentType
-     */
-    public function getContentType()
+    public function getContentType(): ContentType
```
theus77, line 711: Did you try to create a brand new template? I'm afraid that the field migth be null in that case

Davidmattei: Fixed add a giveContentType and reverted this change

src/Command/Notification/BulkActionCommand.php
Davidmattei: Fixed used the count on the elasticSearch service

src/Entity/ContentType.php
Davidmattei: Me too

theus77 (approved):

PR #624, opened 2021-03-20
Feat/default bulk size

theus77:
|Q              |A  |
|---------------|---|
|Bug fix?       |N|
|New feature?   |Y|	
|BC breaks?     |Y|	
|Deprecations?  |Y|	
|Fixed tickets  |Y|	

Typically the bulk size depends of the biggest document's size. Until now you were able to define the default value at command launch. But now you can define it once at the projet level. It avoid to forget to define it again and again.

Also an upcoming monitoring script, which will create missing environment's aliases in elasticsearch, needs this feature.

### DIFF
--- a/src/Command/DocumentCommand.php
+++ b/src/Command/DocumentCommand.php
@@ -39,12 +39,15 @@ class DocumentCommand extends Command
     const ARGUMENT_CONTENTTYPE = 'contentTypeName';
     /** @var string */
     const ARGUMENT_ARCHIVE = 'archive';
+    /** @var int */
+    private $defaultBulkSize;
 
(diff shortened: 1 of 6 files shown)


src/Command/DocumentCommand.php
```diff
@@ -39,12 +39,15 @@ class DocumentCommand extends Command
     const ARGUMENT_CONTENTTYPE = 'contentTypeName';
     /** @var string */
     const ARGUMENT_ARCHIVE = 'archive';
+    /** @var int */
+    private $defaultBulkSize;
```
Davidmattei, line 43:
```suggestion
    private int $defaultBulkSize;
```

Davidmattei (changes requested):

PR #637, opened 2021-04-03
refactor: user views

theus77:
|Q              |A  |
|---------------|---|
|Bug fix?       |Y|
|New feature?   |Y|	
|BC breaks?     |N|	
|Deprecations?  |N|	
|Fixed tickets  |N|	

#Fixes:

- Order on date column
- comma list without ugly space in entyti table

#Refactor

- Hide loop.index in non sortable data table view

#Feature:

- Header with a icon

### DIFF
--- a/src/Controller/UserController.php
+++ b/src/Controller/UserController.php
@@ -67,7 +67,7 @@ class UserController extends AppController
         $createdColumn->setDataLinks(true);
         $table->addColumn('user.index.column.enabled', 'enabled', [true => 'fa fa-check-square-o', false => 'fa fa-square-o']);
         $createdColumn = $table->addColumn('user.index.column.roles', 'roles');
-        $createdColumn->setArrayOfStrings(true);
+        $createdColumn->setClass('');
         $createdColumn = $table->addColumn('user.index.column.lastLogin', 'lastLogin');
         $createdColumn->setDateTimeProperty(true);
 
--- a/src/Controller/UserController.php
+++ b/src/Controller/UserController.php
@@ -57,8 +57,6 @@ class UserController extends AppController
     public function indexAction(Request $request)
     {
         $table = new EntityTable($this->userService);
-
-        $table->addColumn('user.index.column.id', 'id');
         $table->addColumn('user.index.column.username', 'username');
         $table->addColumn('user.index.column.displayname', 'name');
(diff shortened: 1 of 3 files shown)


Davidmattei (approved):

src/Controller/UserController.php
```diff
@@ -57,17 +57,16 @@ public function __construct(LoggerInterface $logger, FormRegistryInterface $form
     public function indexAction(Request $request)
     {
         $table = new EntityTable($this->userService);
-
-        $table->addColumn('user.index.column.id', 'id');
         $table->addColumn('user.index.column.username', 'username');
         $table->addColumn('user.index.column.displayname', 'name');
-        $table->addColumn('user.index.column.email_notification', 'emailNotification', [true => 'fa fa-check-square-o', false => 'fa fa-square-o']);
+        $column = $table->addColumn('user.index.column.email_notification', 'emailNotification', [true => 'fa fa-check-square-o', false => 'fa fa-square-o']);
+        $column->setIconClass('fa fa-bell');
```
dameert, line 63: I see $colum, then $createdColumn, and that one is used multiple times for different colums.

As we are defining stuff here, why not daisy chain the commands? Just as form configuration does?

```suggestion
        $table
            ->addColumn('user.index.column.email_notification', 'emailNotification', [true => 'fa fa-check-square-o', false => 'fa fa-square-o'])
            ->setIconClass('fa fa-bell');
```

src/Form/Data/TableColumn.php
```diff
@@ -16,7 +16,8 @@ final class TableColumn
     private ?string $iconProperty = null;
     private ?bool $dateTimeProperty = null;
     private ?bool $dataLinks = null;
-    private ?bool $arrayOfStrings = null;
+    private string $class = 'nowrap';
```
dameert, line 19: Hard dependency on the styling system that is used.. Maybe use constants and translate them inside the main template of twig?
```suggestion
    private string $class = 'nowrap';
```

src/Controller/UserController.php
theus77: Agree see my coming PR

src/Form/Data/TableColumn.php
theus77: Agree see my coming PR

PR #684, opened 2021-05-27
fix: api upload assets multiple records

Davidmattei:
|Q              |A  |
|---------------|---|
|Bug fix?       |y|
|New feature?   |n|	
|BC breaks?     |n|	
|Deprecations?  |n|	
|Fixed tickets  |n|	

We are now showing the uploaded assets in EMS. If a user already uploaded a asset don't create a new record.


theus77 (commented):
I'll would prefer a cleaning command.

Davidmattei:
Close aggree

PR #713, opened 2021-06-26
feat: emsco document info

theus77:
|Q              |A  |
|---------------|---|
|Bug fix?       |y|
|New feature?   |y|	
|BC breaks?     |n|	
|Deprecations?  |n|	
|Fixed tickets  |n|	

New emsco_document_info twig filter


src/Common/DocumentInfo.php
```diff
@@ -0,0 +1,77 @@
...
+                if ($environmentName === $environment->getName()) {
+                    return $revision;
+                }
+            }
+        }
+
+        return null;
+    }
+
+    public function isAligned(string $environmentName): bool
```
Davidmattei, line 38: In a is Aligned I was expecting a Hash check ? endTime null only means the currentRevision

Davidmattei (changes requested):

Davidmattei (approved):

PR #718, opened 2021-07-08
fix: form submission

theus77:
|Q              |A  |
|---------------|---|
|Bug fix?       |y|
|New feature?   |n|	
|BC breaks?     |n|	
|Deprecations?  |n|	
|Fixed tickets  |n|	

- expire_date can be null in datamodel
- deadline_date is depreacetd

### DIFF
--- a/src/Service/Form/Submission/FormSubmissionRequest.php
+++ b/src/Service/Form/Submission/FormSubmissionRequest.php
@@ -41,7 +41,7 @@ final class FormSubmissionRequest
         $this->data = $submit['data'];
         $this->files = $submit['files'];
         $this->label = $submit['label'] ?? '';
-        $formattedDate = \DateTime::createFromFormat('c', $submit['expire_date']);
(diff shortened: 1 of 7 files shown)


src/Service/Form/Submission/FormSubmissionRequest.php
```diff
@@ -41,7 +41,7 @@ public function __construct(Request $request)
         $this->data = $submit['data'];
         $this->files = $submit['files'];
         $this->label = $submit['label'] ?? '';
-        $formattedDate = \DateTime::createFromFormat('c', $submit['expire_date']);
+        $formattedDate = \DateTime::createFromFormat(\DateTime::ISO8601, $submit['expire_date']);
```
Davidmattei, line 44: Note: This format is not compatible with ISO-8601, but is left this way for backward compatibility reasons. Use DateTime::ATOM or DATE_ATOM for compatibility with ISO-8601 instead.

Davidmattei (changes requested):
One small change request

PR #752, opened 2021-08-24
fix: content transformers command

Davidmattei:
|Q              |A  |
|---------------|---|
|Bug fix?       |y|
|New feature?   |n|	
|BC breaks?     |n|	
|Deprecations?  |n|	
|Fixed tickets  |n|	

@theus77 in the dataService -> lockRevision, added a null === $username, because the appTwig inMyCircles only works for authenticated users. When we pass a username we are running from cli. Aggree?


coppee (approved):

theus77:
I agree.

Davidmattei:
> I agree.

👍

PR #756, opened 2021-08-24
fix: nested files collection not working

Davidmattei:
|Q              |A  |
|---------------|---|
|Bug fix?       |y|
|New feature?   |n|	
|BC breaks?     |n|	
|Deprecations?  |n|	
|Fixed tickets  |n|	

For nested collections do not add the key.

Error in twig:
Key "0" in object with ArrayAccess of class "Doctrine\Common\Collections\ArrayCollection" does not exist.


Davidmattei:
It was just wrong data in a specific project, used the renumbering option for fixing this.

PR #765, opened 2021-09-01
fix: dompdf chroot

theus77:
|Q              |A  |
|---------------|---|
|Bug fix?       |y|
|New feature?   |y|	
|BC breaks?     |n|	
|Deprecations?  |n|	
|Fixed tickets  |n|	

With new version of DomPdf the DomPdf option must be specify in order to allow access to assets like jpeg or png. This has been fixed in the common DomPdfPrinter. So it's time to refactor PDF action to use the Common DomPdfPrinter.


Davidmattei (approved):
Approved, but I would freeze the DataController and move actions to dedicated slim controllers.

PR #768, opened 2021-09-02
fix: job commands

Davidmattei:
|Q              |A  |
|---------------|---|
|Bug fix?       |y|
|New feature?   |n|	
|BC breaks?     |n|	
|Deprecations?  |n|	
|Fixed tickets  |n|	

- fix: align command update snapshot
> Because the align command uses a transaction -> doctrine clear. We need to do a new findByName before updating the snapshot environment.

- fix: job with sub commands 
> The job output and finish should fetch the job entity again. Because the command or multiple sub commands can clear doctrine for performance reasons.

### DIFF
--- a/src/Repository/JobRepository.php
+++ b/src/Repository/JobRepository.php
@@ -11,13 +11,7 @@ class JobRepository extends EntityRepository
 {
     public function findById(int $jobId): Job
     {
-        $qb = $this->createQueryBuilder('j');
-        $query = $qb
-            ->andWhere($qb->expr()->eq('j.id', ':job_id'))
-            ->setParameter('job_id', $jobId)
-            ->getQuery();
-
-        $job = $query->getSingleResult();
+        $job = $this->findOneBy(['id' => $jobId]);
 
         if (!$job instanceof Job) {
             throw new \RuntimeException('Job not found');
--- a/src/Repository/JobRepository.php
+++ b/src/Repository/JobRepository.php
@@ -5,9 +5,27 @@ declare(strict_types=1);
 namespace EMS\CoreBundle\Repository;
 
 use Doctrine\ORM\EntityRepository;
+use EMS\CoreBundle\Entity\Job;
 
 class JobRepository extends EntityRepository
 {
+    public function findById(int $jobId): Job
+    {
+        $qb = $this->createQueryBuilder('j');
+        $query = $qb
+            ->andWhere($qb->expr()->eq('j.id', ':job_id'))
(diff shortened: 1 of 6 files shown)


src/Repository/JobRepository.php
```diff
@@ -5,9 +5,27 @@
...
 {
+    public function findById(int $jobId): Job
+    {
+        $qb = $this->createQueryBuilder('j');
+        $query = $qb
+            ->andWhere($qb->expr()->eq('j.id', ':job_id'))
+            ->setParameter('job_id', $jobId)
+            ->getQuery();
+
+        $job = $query->getSingleResult();
```
theus77, line 20: As you search for by the id, find should works:
```suggestion
        $job = $this->find($jobId);
```

src/Service/EnvironmentService.php
```diff
@@ -63,6 +65,12 @@ public function __construct(
         $this->logger = $logger;
         $this->elasticaService = $elasticaService;
         $this->instanceId = $instanceId;
+
+        $environmentRepository = $doctrine->getRepository(Environment::class);
+        if (!$environmentRepository instanceof EnvironmentRepository) {
+            throw new \RuntimeException('Not found repository');
+        }
+        $this->environmentRepository = $environmentRepository;
```
theus77, line 73: I don't like code in service constructors. It may impact the overall performances as they are called on each request. Personally, I would prefer a getter with a singleton. Or an autowiring stuff

```
        private function getEnvironmentRepository(): EnvironmentRepository
        {
                if (null !== $this->environmentRepository) {
                    return $this->environmentRepository;
                }
                $environmentRepository = $this->doctrine->getRepository(Environment::class);
                if (!$environmentRepository instanceof EnvironmentRepository) {
                    throw new \RuntimeException('Not found repository');
                }
                $this->environmentRepository = $environmentRepository;
                return $this->environmentRepository;
        }
```

src/Repository/JobRepository.php
Davidmattei: aggree

src/Service/EnvironmentService.php
Davidmattei: Should be fixed if we make a service for the Environment repository, but out of scope for this pr

src/Repository/JobRepository.php
Davidmattei: https://github.com/ems-project/EMSCoreBundle/pull/768/commits/74c4af8502ce1086e69f300ec66437053e548263

PR #786, opened 2021-09-10
refactor: revision data

theus77:
|Q              |A  |
|---------------|---|
|Bug fix?       |n|
|New feature?   |y|	
|BC breaks?     |n|	
|Deprecations?  |n|	
|Fixed tickets  |n|	

migration to emsco-twigs translation domain and refactor revision toolbar with dropdown menus


coppee:
@theus77 @dameert  @Davidmattei  Tricky question: from a User/Customer point of view, wouldn't it be more obvious to use the term **version** instead of **revision**?

theus77:
> @theus77 @dameert  @Davidmattei  Tricky question: from a User/Customer point of view, wouldn't it be more obvious to use the term **version** instead of **revision**? 

In fact "version" terminology also exists with the versioning mechanism allowing to have forked version of a document within the same environment

dameert:
> > @theus77 @dameert  @Davidmattei  Tricky question: from a User/Customer point of view, wouldn't it be more obvious to use the term **version** instead of **revision**? 
> 
> In fact "version" terminology also exists with the versioning mechanism allowing to have forked version of a document within the same environment

But other CMS systems also use the term Revision.

dameert:
> @theus77 @dameert @Davidmattei Tricky question: from a User/Customer point of view, wouldn't it be more obvious to use the term **version** instead of **revision**?

To make a revision of something that is written or something that has been decided means to make changes to it in order to improve it, make it more modern, or make it more suitable for a particular purpose. The phase of writing that is actually most important is revision

coppee:
Thanks for the feedback, personally I agree with you (except a lot of CMS are using Version in this case).  But from a CX perspective, I think it's not entirely obvious.  I challenge you (or the webmaster) to explain the concept of revision to a client without using the term version.  :)

Davidmattei (approved):

PR #789, opened 2021-09-14
refactor: draft in progress

theus77:
|Q              |A  |
|---------------|---|
|Bug fix?       |n|
|New feature?   |y|	
|BC breaks?     |n|	
|Deprecations?  |n|	
|Fixed tickets  |n|	

Migrate the draft in progress to datable  and renew buttons


src/Routes.php
```diff
@@ -0,0 +1,12 @@
...
+
+namespace EMS\CoreBundle;
+
+class Routes
+{
+    public const EditRevision = 'emsco_edit_revision';
+    public const ViewRevision = 'emsco_view_revisions';
+    public const DiscardDraft = 'emsco_discard_draft';
+    public const DraftInProgress = 'emsco_draft_in_progress';
+    public const DraftInProgressAjax = 'emsco_draft_in_progress_ajax';
```
Davidmattei, line 11: const should always by uppercase

```suggestion
    public const EDIT_REVISION = 'emsco_edit_revision';
    public const VIEW_REVISION = 'emsco_view_revisions';
    ...
```

Davidmattei (changes requested):
In the EMSCoreExtension::prepend function we are setting twig globals, maybe we should also pass the new Routes class, so we can use those constants in twig?

Would also name the consts differently EMSCO_REVISON_(EDIT,VIEW,....) instead of EMSCO_EDIT_REVISION.

Davidmattei (approved):

PR #821, opened 2021-10-30
fix: ordered static data link

theus77:
|Q              |A  |
|---------------|---|
|Bug fix?       |y|
|New feature?   |n|	
|BC breaks?     |n|	
|Deprecations?  |n|	
|Fixed tickets  |n|	

Bulk query can't be ordered (the order is defined by the cluster) Or at least it's a limitation of using a elastica client 5 with a elasticsearch 7. Never the less, this assume that loading the first 1000 results in a static data link is enough (you can still switch to a dynamic data link


Davidmattei:
Already released https://github.com/ems-project/EMSCoreBundle/pull/819

PR #847, opened 2021-11-21
feat: schedule

theus77:
|Q              |A  |
|---------------|---|
|Bug fix?       |n|
|New feature?   |y|	
|BC breaks?     |y|	
|Deprecations?  |n|	
|Fixed tickets  |n|	

Job scheduler (CRUD + Job)

## Breaking change
The job emsco:release:publish is not automatically started every seconds

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,6 +31,7 @@
 		"doctrine/doctrine-migrations-bundle": "^1.3",
 		"doctrine/orm": "^2.7",
 		"doctrine/persistence": "^1.3",
+		"dragonmantank/cron-expression": "^3.1",
 		"elasticms/client-helper-bundle": "^3.6",
 		"guzzlehttp/guzzle": "^6.3",
 		"ocramius/doctrine-batch-utils": "2.1",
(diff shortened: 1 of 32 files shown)


composer.json
```diff
@@ -56,8 +57,7 @@
 		"symfony/validator": "^4.4",
 		"symfony/web-link": "^4.4",
 		"symfony/webpack-encore-bundle": "^1.0",
-		"symfony/yaml": "^4.4",
-		"ramsey/uuid-doctrine": "^1.6"
```
Davidmattei, line 60: why is ramsey/uuid removed?

src/Form/Form/ScheduleType.php
```diff
@@ -0,0 +1,81 @@
+<?php
+
```
Davidmattei, line 2:
```suggestion
declare(strict_types=1);
```

Davidmattei (changes requested):

composer.json
theus77: It was there twice

src/Form/Form/ScheduleType.php
theus77: Done: 558ad6dae7580f4fd384745420ee96e9581aaf55

Davidmattei (approved):

PR #874, opened 2021-12-08
fix: data locale attribute for objectpicker

Davidmattei:
|Q              |A  |
|---------------|---|
|Bug fix?       |n|
|New feature?   |y|	
|BC breaks?     |n|	
|Deprecations?  |n|	
|Fixed tickets  |n|	

Make it possible to set the locale attribute on a object picker.
For example using the following html code in a custom view

```html
<select id="select-page" 
    name="select-page" 
    data-type="page"
    data-locale="nl"
    data-dynamic-loading="data-dynamic-loading"
    class="objectpicker form-control ignore-ems-update">
</select>
```


theus77 (approved):
Don't forget to regenerate the asssets once all PR merged

PR #888, opened 2021-12-19
refactor: form entities from submission

theus77:
|Q              |A  |
|---------------|---|
|Bug fix?       |y|
|New feature?   |n|	
|BC breaks?     |n|	
|Deprecations?  |n|	
|Fixed tickets  |n|	

With the integration of the submission bundle in elasticms we are facing conflict between the Core's FormSubmission entities and the Submission's one. This PR is about adding a dependency to SubmissionBundle and use Submission's entities. 
Fix the doctrine:migrate:diff command


theus77:
I have tried to have inheritance from Submission entities (but it's really complex). So the less code and pragmatic approach wins

src/Form/Data/EntityRow.php
```diff
@@ -4,7 +4,7 @@
 
 namespace EMS\CoreBundle\Form\Data;
 
-use EMS\CoreBundle\Entity\EntityInterface;
+use EMS\SubmissionBundle\Entity\EntityInterface;
```
Davidmattei, line 7: I do not like this, this is the EntityRow for all entity datatables. And It now uses the SubmissionBundle entityInterface allways?

Davidmattei (changes requested):

src/Form/Data/EntityRow.php
Davidmattei: And the src/Entity/EntityInterface.php  is extension of EMS\SubmissionBundle\Entity\EntityInterface

theus77: It extends a common interface now

Davidmattei (approved):

PR #946, opened 2022-02-05
feat: admin content type api

theus77:
|Q              |A  |
|---------------|---|
|Bug fix?       |n|
|New feature?   |y|	
|BC breaks?     |n|	
|Deprecations?  |n|	
|Fixed tickets  |n|	

Add a content-type api


src/DependencyInjection/Compiler/EntityServiceCompilerPass.php
```diff
@@ -0,0 +1,32 @@
+<?php
```
Davidmattei, line 1: If we just want to injected tagged services we do not need a custom CompilerPass

https://symfony.com/doc/4.4/service_container/tags.html#reference-tagged-services

theus77: Tks, I learned something

Davidmattei (approved):

PR #962, opened 2022-02-15
fix: issue with tabs in collections

theus77:
|Q              |A  |
|---------------|---|
|Bug fix?       |y|
|New feature?   |n|	
|BC breaks?     |n|	
|Deprecations?  |n|	
|Fixed tickets  |n|	

New path parameter that is transmitted recursively

### DIFF
--- a/src/Resources/views/macros/data-field-type.html.twig
+++ b/src/Resources/views/macros/data-field-type.html.twig
@@ -5,7 +5,7 @@
 	{{ _self.messages(dataField) }}
 {% endmacro %}
 
-{% macro renderDataFieldBlock(dataField, source, compare, compareRawData, blockName = '') %}
+{% macro renderDataFieldBlock(dataField, source, compare, compareRawData, blockName = '', path = []) %}
 	{% if blockName is same as('') %}
 		{% set blockName = dataField.fieldType.type|replace({'\\': '_'}) %}
 	{% endif %}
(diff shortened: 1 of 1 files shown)


src/Resources/views/macros/data-field-type.html.twig
```diff
@@ -392,19 +394,19 @@
 		<ul class="nav nav-tabs">
 			{% for value in values|default([]) %}
 				<li class="{% if loop.index == 1%}active{% endif %}">
-					<a href="#item_{{ dataField.fieldType.id~'_'~loop.index }}__tab" data-toggle="tab" aria-expanded="{% if loop.index == 1%}true{% else %}false{% endif %}">
+					<a href="#item_{{ path|join('_')~'_'~loop.index0 }}__tab" data-toggle="tab" aria-expanded="{% if loop.index == 1%}true{% else %}false{% endif %}">
```
Davidmattei, line 397: index0 ?

Davidmattei (changes requested):
index0 is that working?

src/Resources/views/macros/data-field-type.html.twig
Davidmattei: Good! Yes I learned something

loop.index0 | The current iteration of the loop starting at 0.

Davidmattei (approved):

PR #964, opened 2022-02-15
feat: analyzer admin api

theus77:
|Q              |A  |
|---------------|---|
|Bug fix?       |n|
|New feature?   |y|	
|BC breaks?     |n|	
|Deprecations?  |n|	
|Fixed tickets  |n|

### DIFF
--- a/src/Core/Mapping/AnalyzerManager.php
+++ b/src/Core/Mapping/AnalyzerManager.php
@@ -9,11 +9,11 @@ use EMS\CoreBundle\Service\EntityServiceInterface;
 
 class AnalyzerManager implements EntityServiceInterface
 {
-    private AnalyzerRepository $AnalyzerRepository;
+    private AnalyzerRepository $analyzerRepository;
 
-    public function __construct(AnalyzerRepository $AnalyzerRepository)
+    public function __construct(AnalyzerRepository $analyzerRepository)
     {
-        $this->AnalyzerRepository = $AnalyzerRepository;
+        $this->analyzerRepository = $analyzerRepository;
     }
 
     public function isSortable(): bool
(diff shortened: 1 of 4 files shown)


src/Core/Mapping/AnalyzerManager.php
```diff
@@ -0,0 +1,87 @@
...
+
+    public function getEntityName(): string
+    {
+        return 'analyzer';
+    }
+
+    public function getAliasesName(): array
+    {
+        return [
+            'analyzers',
```
AdrienHq, line 37: 'analyzer' is missing ?

src/Core/Mapping/AnalyzerManager.php
```diff
@@ -0,0 +1,87 @@
...
+namespace EMS\CoreBundle\Core\Mapping;
+
+use EMS\CommonBundle\Entity\EntityInterface;
+use EMS\CoreBundle\Entity\Analyzer;
+use EMS\CoreBundle\Repository\AnalyzerRepository;
+use EMS\CoreBundle\Service\EntityServiceInterface;
+
+class AnalyzerManager implements EntityServiceInterface
+{
+    private AnalyzerRepository $AnalyzerRepository;
```
Davidmattei, line 12:
```suggestion
    private AnalyzerRepository $analyzerRepository;
```

Davidmattei: And change in this class $this->AnalyzerRepositoryto $this->analyzerRepository

Davidmattei (changes requested):

src/Core/Mapping/AnalyzerManager.php
theus77: It's in the getEntityName
```
    public function getEntityName(): string
    {
        return 'analyzer';
    }
```

Davidmattei (approved):

PR #995, opened 2022-03-02
fix: remove old subtitle

Davidmattei:
|Q              |A  |
|---------------|---|
|Bug fix?       |y|
|New feature?   |n|	
|BC breaks?     |n|	
|Deprecations?  |n|	
|Fixed tickets  |n|


Davidmattei:
Maybe in the feature we need an i18n key for this, but for now I would just remove it

theus77 (approved):

PR #1003, opened 2022-03-06
feat: i18n twig filter take into account a fallback locale

theus77:
|Q              |A  |
|---------------|---|
|Bug fix?       |n|
|New feature?   |y|	
|BC breaks?     |n|	
|Deprecations?  |n|	
|Fixed tickets  |n|	

feat: i18n twig filter take into account a fallback locale


Davidmattei (changes requested):
👍

Davidmattei (approved):

PR #1087, opened 2022-05-12
feat: datalinks referrer document

Davidmattei:
|Q              |A  |
|---------------|---|
|Bug fix?       |y|
|New feature?   |y|	
|BC breaks?     |n|	
|Deprecations?  |n|	
|Fixed tickets  |n|	

In the custom dataLinks view type we need to have the referrer document. 
So that we can make the choices context aware.

Tested: 
- Custom view searches
- Query searches
- Deprecated searches
- DataLink in collections


theus77 (approved):
A new 1.20.0 release

PR #1092, opened 2022-05-17
fix: copy revision command broken

Davidmattei:
|Q              |A  |
|---------------|---|
|Bug fix?       |y|
|New feature?   |n|	
|BC breaks?     |y|	
|Deprecations?  |n|	
|Fixed tickets  |n|	

The bc break is the argument searchQuery before we passed: 

> {\"query\":{\"bool\":{\"must\":[{\"term\":{\"_contenttype\":{\"value\":\"publication\"}}},{\"term\":{\"month\":{\"value\":\"04\/2022\"}}}]}}}

Now we suppress the query like other commands:

> :{\"bool\":{\"must\":[{\"term\":{\"_contenttype\":{\"value\":\"publication\"}}},{\"term\":{\"month\":{\"value\":\"04\/2022\"}}}]}}

Now that we have the revisionSearcher, we do not need that many logic for coping revisions.


coppee (approved):

IsaMic (approved):

src/Service/Revision/RevisionService.php
```diff
@@ -173,6 +173,22 @@ public function create(ContentType $contentType, ?UuidInterface $uuid = null, ar
...
+     */
+    public function copy(Revision $revision, ?array $mergeRawData = null): void
+    {
+        $copiedRevision = $revision->clone();
+
+        if ($mergeRawData) {
+            $copiedRevision->setRawData(\array_merge($copiedRevision->getRawData(), $mergeRawData));
+        }
+
+        $form = null;
```
vfauvarque, line 187: Why not pass null directly to the method below, instead of instantiating a variable ?

vfauvarque (approved):

src/Service/Revision/RevisionService.php
Davidmattei: Aah go reviewing, because the finalizeDraft has &$form as input argument. It will be refactored in the future. For now it's to make php happy.

vfauvarque: fair enough

PR #1143, opened 2022-07-11
fix: update referrers

Davidmattei:
|Q              |A  |
|---------------|---|
|Bug fix?       |y|
|New feature?   |n|	
|BC breaks?     |n|	
|Deprecations?  |n|	
|Fixed tickets  |n|	

# Update referrers

If a data link field type has the extra option 'update referrers field' defined, then on post finalization an event will be dispatched.
They event listener will search the linked document and update the referrers field.

## Example: 
- **Club** content type with a multiple data link field called **players**. 
- **Player** content type with a single or multiple data link field type called club(s) and a update referrers value **players**
If we create or update a player and we change the club, the referred club field **players** will be updated.

## Refactor
- Moved to the update referrer logic to a new revision event subscriber.
- Duplicated for loop logic (add/remove) moved to simple callback
- Changed signature postFinalizeTreatment passing revision instead of string type and ouuid

# Status
We are not quite happy, because it will create a lot of revisions. Maybe we should use the reindex instead of creating revisions

### DIFF
--- a/src/Core/Revision/Json/JsonMenuNestedDefinition.php
+++ b/src/Core/Revision/Json/JsonMenuNestedDefinition.php
@@ -333,7 +333,7 @@ final class JsonMenuNestedDefinition
     }
 
     /**
-     * @param Collection<string, FieldType> $children
+     * @param Collection<int, FieldType> $children
      */
     private function isGrantedForAllDescendant(Collection $children): bool
     {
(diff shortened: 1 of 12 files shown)


Davidmattei:
https://github.com/ems-project/elasticms/issues/28

PR #1152, opened 2022-07-25
fix: save/edit publish revision

Davidmattei:
|Q              |A  |
|---------------|---|
|Bug fix?       |y|
|New feature?   |n|	
|BC breaks?     |n|	
|Deprecations?  |n|	
|Fixed tickets  |n|	

* Edit a revision that is not published in any environments
    * Showing a blue submit button"Save and publish in" >  fixed now showing blue "Save" button. Because we will only update the revision in the database.
* Edit archived revision that is published in live environment
    * Showing a gray submit button "Save as draft" > fixed now showing blue "Save and publish in Live"
    
On save we only published if they revision had an endTime, this check could be removed. If we save a revision and it has environments, we should publish.

Creating a draft -> new revision (no environments), so will not conflict with above logic.

### DIFF
--- a/src/Controller/Revision/EditController.php
+++ b/src/Controller/Revision/EditController.php
@@ -182,10 +182,8 @@ class EditController extends AbstractController
                     $revision = $this->revisionService->saveVersion($revision, $objectArray, $versionTag, $form);
                 } else {
                     $this->revisionService->save($revision, $objectArray);
-                    if (null !== $revision->getEndTime()) {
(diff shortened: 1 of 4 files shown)


theus77 (approved):

Davidmattei:
Improve, check that we can not do the "edit revison" on the current revision in the default environment. 
Because then we allow changes to the document that are not saved in the db

theus77 (approved):

PR #1158, opened 2022-07-27
fix: reload data reindex

Davidmattei:
|Q              |A  |
|---------------|---|
|Bug fix?       |y|
|New feature?   |n|	
|BC breaks?     |n|	
|Deprecations?  |n|	
|Fixed tickets  |n|	

The rawData gets recomputed/reloaded and indexed in environment, but we are not saving it to the database.

On reindex the postprocessing was not passing finalized and migration false.

New selfUpdate property on Revision: for allowing revision sql updates without locking

### DIFF
--- a/src/Command/ReindexCommand.php
+++ b/src/Command/ReindexCommand.php
@@ -36,12 +36,10 @@ class ReindexCommand extends EmsCommand
     protected $dataService;
     /** @var string */
     private $instanceId;
-    /** @var int */
-    private $count;
-    /** @var int */
-    private $deleted;
-    /** @var int */
-    private $error;
+    private int $count = 0;
+    private int $deleted = 0;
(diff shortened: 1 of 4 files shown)


src/Command/ReindexCommand.php
```diff
@@ -194,7 +188,8 @@ public function reindex(string $name, ContentType $contentType, ?string $index,
                         ]);
                     } else {
                         if ($reloadData) {
-                            $this->dataService->reloadData($revision);
+                            $reloadedResult = $this->dataService->reloadData($revision, false);
+                            $this->reloaded += $reloadedResult;
```
theus77, line 192:
```suggestion
                            $this->reloaded += $this->dataService->reloadData($revision, false);
```

theus77 (dismissed):

PR #1170, opened 2022-08-17
fix: json menu nested descendant grant

Davidmattei:
|Q              |A  |
|---------------|---|
|Bug fix?       |y|
|New feature?   |y|	
|BC breaks?     |y|	
|Deprecations?  |n|	
|Fixed tickets  |n|	

User should be able to add item of type even if they are not granted for adding other item types.
Also we can deny some fields for the user, but still allow edit/add on the item.

New feature, block template for overriding root actions

https://github.com/ems-project/EMSCommonBundle/pull/442/files

### DIFF
--- a/doc/twig/jsonMenuNested.md
+++ b/doc/twig/jsonMenuNested.md
@@ -98,12 +98,27 @@ In the html string you can access the option **context** and:
 ```twig
     {% set context = { 'title': 'Example' } %}
     {% set pageButtons %}
-        {{ '<button class="btn-test btn btn-sm btn-primary" data-item-id="{{ item.id }}">{{ item }}</button>' }}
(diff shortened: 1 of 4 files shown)


Davidmattei:
The item_type -> 'root' is not correct, this means we can not create a child with type root.
It should be *_json_root*, the underscore prefix means it's an internal defined field

Davidmattei:
> The item_type -> 'root' is not correct, this means we can not create a child with type root. It should be __json_root_, the underscore prefix means it's an internal defined field

Fixed 
https://github.com/ems-project/EMSCommonBundle/pull/442/files
https://github.com/ems-project/EMSCoreBundle/pull/1170/commits/2708e1316387262c6046a30093112e6ce617a4c3

coppee (approved):

PR #1175, opened 2022-08-23
fix: do not overwrite generateUrl in controller

Davidmattei:
|Q              |A  |
|---------------|---|
|Bug fix?       |y|
|New feature?   |n|	
|BC breaks?     |n|	
|Deprecations?  |n|	
|Fixed tickets  |n|	

Better solution for fix in https://github.com/ems-project/EMSCoreBundle/pull/1094
Tested browse from server.


coppee (commented):
Reviewed

PR #1220, opened 2022-10-18
feat: treat accept reject disabled

IsaMic:
|Q              |A  |
|---------------|---|
|Bug fix?       | n|
|New feature?   |y|	
|BC breaks?     |n|	
|Deprecations?  |n|	
|Fixed tickets  |n|	

Treat Notifications :  If no value for field publishTo only reject button is available, and I a value exist in publishTo only accept button is available.


Davidmattei (approved):
👍

PR #1227, opened 2022-10-21
fix: change _self target links to _parent in iframe preview

coppee:
|Q              |A  |
|---------------|---|
|Bug fix?       |n|
|New feature?   |y|	
|BC breaks?     |n|	
|Deprecations?  |n|	
|Fixed tickets  |n|	

To avoid to load the link in the iframe


assets/js/module/iframePreview.js
```diff
@@ -63,4 +64,16 @@ export default class IframePreview {
             self.loadIframes(iframes)
         });
     }
+
+    #changeSelfTargetLinksToParent(body) {
```
theus77, line 68: Why a #? Does that mean something specific?

assets/js/module/iframePreview.js
```diff
@@ -63,4 +64,16 @@ export default class IframePreview {
...
+        const parser = new DOMParser()
+        const dom = parser.parseFromString(body, 'text/html')
+        ;[...dom.getElementsByTagName('a')].forEach((link) => {
+            if (!link.getAttribute('target')) {
+                link.setAttribute('target', '_parent')
+            }
+        })
+
+        return dom.documentElement.outerHTML
```
theus77, line 77: I feel like I'll lean a lot why you. :-). Why no semi-colon at the end of lines?

What if a internal link has a target? Nevertheless it's good for a first implementation. Will see how it works

theus77 (approved):

assets/js/module/iframePreview.js
coppee: It specify a private method. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields

https://caniuse.com/mdn-javascript_classes_private_class_methods

assets/js/module/iframePreview.js
coppee: - It's an habit but it's a good point, we need to specify our JS dev guidelines

- I was thinking that if target is declared  it will be with _blank value and make sense for me to keep the behavior. But we can talk about and I will make also a more precise condition

theus77: Good idea, let's try eslint? https://eslint.org/

Davidmattei: 👍

Davidmattei (approved):

PR #1240, opened 2022-10-31
feat(revision-version): make version on publication

Davidmattei:
|Q              |A  |
|---------------|---|
|Bug fix?       |n|
|New feature?   |y|	
|BC breaks?     |n|	
|Deprecations?  |n|	
|Fixed tickets  |n|


src/Entity/Revision.php
```diff
@@ -931,9 +944,9 @@ public function getVersionDate(string $field): ?\DateTimeImmutable
             return null;
         }
 
-        $dateTime = \DateTimeImmutable::createFromFormat(\DateTimeImmutable::ATOM, $dateString);
+        $dateTime = \DateTimeImmutable::createFromFormat(\DateTimeInterface::ATOM, $dateString);
 
-        return $dateTime ? $dateTime : null;
+        return $dateTime ?: null;
```
theus77:
```suggestion
        return false === $dateTime ? null :  $dateTime;
```

theus77 (approved):

PR #1244, opened 2022-11-04
fix(phpstan): Comparison always true

coppee:
Comparison operation ">" between int<1, max> and 0 is always true.

|Q              |A  |
|---------------|---|
|Bug fix?       |Y|
|New feature?   |N|	
|BC breaks?     |N|	
|Deprecations?  |N|	
|Fixed tickets  |N|


theus77 (dismissed):

Davidmattei:
Close fixed in https://github.com/ems-project/EMSCoreBundle/pull/1245

PR #1246, opened 2022-11-07
feat(environment): template publication

Davidmattei:
|Q              |A  |
|---------------|---|
|Bug fix?       |n|
|New feature?   |y|	
|BC breaks?     |n|	
|Deprecations?  |n|	
|Fixed tickets  |n|	

Add the possibility to define a publication template on an environment.
If defined we should check the publication template before publication.

Created a EnvironmentPublisher service with his own factory.
The factory will remember previous publishers, for using it during a bulk publication

https://ems-project.github.io/#/./elasticms/environment/environment


coppee (dismissed):

src/Form/Form/EditEnvironmentType.php
```diff
@@ -22,51 +23,62 @@ class EditEnvironmentType extends AbstractType
...
-                        'class' => 'btn btn-primary btn-sm ',
-                ],
-                'icon' => 'fa fa-save',
-        ]);
     }
 
     public function configureOptions(OptionsResolver $resolver): void
     {
-        /* set the default option value for this kind of compound field */
-        parent::configureOptions($resolver);
```
theus77, line 69: Are you sure that it's not required to call the parent's configureOptions fucntion?

src/Service/Revision/RevisionService.php
```diff
@@ -186,6 +186,11 @@ public function getDocumentInfo(EMSLink $documentLink): DocumentInfo
         return new DocumentInfo($documentLink, $this->revisionRepository->findAllPublishedRevision($documentLink));
     }
 
+    public function getDocumentInfoByRevision(Revision $revision): DocumentInfo
```
theus77: Does the DocumentInfo entity is abel to know if a revision can be published in a specific environment (for overview tables)?

Davidmattei: Good remark this change has been removed, did not need the documentInfo but the document.

If you want the documentInfo in the publicaton template -> revison.emsLink|emsco_document_info

src/Form/Form/EditEnvironmentType.php
Davidmattei: yes, the parent class is Symfony\Component\Form\AbstractType; which has no implementation on configureOptions.

coppee (approved):

